PR #55, opened 2022-07-07
Bnb/pipeline

bnb32:


sup3r/cli.py
```diff
@@ -28,26 +32,86 @@ def main(ctx, config_file, verbose):
     ctx.obj['VERBOSE'] = verbose
 
 
-@main.group()
+@main.group(invoke_without_command=True)
```
grantbuster, line 35: You should look up click groups vs. commands. Any function like forward_pass that doesnt have any follow on commands should be the final command and not be a group. You only need these to be groups if there are follow on commands like the valid config keys thing

sup3r/cli.py
```diff
@@ -28,26 +32,86 @@ def main(ctx, config_file, verbose):
...
+        config_file = ctx.obj['CONFIG_FILE']
+        verbose = any([verbose, ctx.obj['VERBOSE']])
+        ctx.invoke(dc_cli, config_file=config_file, verbose=verbose)
+
+
+@main.group(invoke_without_command=True)
+@click.option('-v', '--verbose', is_flag=True,
+              help='Flag to turn on debug logging.')
+@click.pass_context
+def data_collect(ctx, verbose):
```
grantbuster, line 75: duplicate?

sup3r/pipeline/forward_pass.py
```diff
@@ -719,13 +720,25 @@ def get_node_cmd(cls, config):
...
         cmd = (f"python -c \'{import_str};\n"
                f"logger = init_logger({log_arg_str});\n"
                f"strategy = {fps_init_str};\n"
                f"fwp = {cls.__name__}({fwp_arg_str});\n"
-               "fwp.run()\'\n").replace('\\', '/')
+               "fwp.run()")
+
+        if job_name is not None:
+            cmd += (";\nfrom reV.pipeline.status import Status;\n"
+                    f"Status.make_job_file({status_file_arg_str})")
```
grantbuster, line 738: Nice looks good!

tests/test_pipeline.py
```diff
@@ -0,0 +1,90 @@
...
+        Pipeline.run(tmp_fpipeline, monitor=True)
+
+        assert os.path.exists(fp_out)
+
+        status_file = glob.glob(os.path.join(td, '*_status.json'))[0]
+        with open(status_file, 'r') as fh:
+            status = json.load(fh)
+            assert all(s not in str(status)
+                       for s in ('fail', 'pending', 'submitted'))
+            assert 'successful' in str(status)
```
grantbuster, line 90: Maybe check that multiple modules were called too?

grantbuster (changes requested):

sup3r/cli.py
grantbuster: @bnb32 one more thing on this, making the commands into commands instead of groups i dont think you need the lines that say `if ctx.invoked_subcommand is None:` because there wont be an option for subcommand.

bnb32: yeah good point

grantbuster (approved):

PR #58, opened 2022-07-18
Bnb/eval

bnb32:
Mostly refactoring things. Only new features: getting full domain for wrf data if target, shape, and raster_file are None and config file generation in SUP3R.init_pass_collect.


sup3r/pipeline/forward_pass.py
```diff
@@ -746,65 +812,77 @@ def run(self):
...
-                                                      self.hr_crop_slices):
-
-                data_chunk = self.data_handler.data[s_low_pad]
-                data[s_high] = ForwardPass.forward_pass_chunk(
-                    data_chunk, crop_slices=s_high_crop,
-                    model_path=self.model_path)
+        if self.cache_file is not None and os.path.exists(self.cache_file):
+            logger.info('Loading forward pass output from cache file: '
+                        f'{self.cache_file}')
+            with open(self.cache_file, 'rb') as f:
```
grantbuster, line 818: Caching the output in a pkl? This seems unnecessary... Isnt the output saved to h5/nc? I feel like we should assume that the forward pass always needs to be run unless the pipeline/status module has a successful flag.

sup3r/pipeline/forward_pass.py
```diff
@@ -746,65 +812,77 @@ def run(self):
...
+                   f'{self.strategy.spatial_overlap} and temporal_overlap of '
+                   f'{self.strategy.temporal_overlap}.')
+            logger.info(msg)
+            max_workers = self.max_workers
+            data = np.zeros((self.strategy.s_enhance * self.data_shape[0],
+                             self.strategy.s_enhance * self.data_shape[1],
+                             self.strategy.t_enhance * self.data_shape[2],
+                             len(self.output_features)),
+                            dtype=np.float32)
+            if max_workers == 1:
```
grantbuster, line 833: This method is wayyy too complex. Consider breaking up into _run_serial() and _run_parallel()

sup3r/pipeline/forward_pass_cli.py
```diff
@@ -68,21 +68,30 @@ def from_config(ctx, config_file, verbose):
...
     logger.debug('Hardware run option: "{}"'.format(hardware_option))
 
     sig = signature(ForwardPassStrategy)
     strategy_kwargs = {k: v for k, v in config.items()
                        if k in sig.parameters.keys()}
     strategy = ForwardPassStrategy(**strategy_kwargs)
 
-    for i in range(strategy.nodes):
+    node_index = config.get('node_index', None)
+    nodes = [node_index] if node_index is not None else range(strategy.nodes)
```
grantbuster, line 88: I'm not sure if this is the right place but can we str(node_index).zfill(5) somewhere?

sup3r/preprocessing/batch_handling.py
```diff
@@ -427,24 +481,19 @@ def parallel_normalization(self, max_workers=None):
                             f'data handlers in {dt.now() - now}.')
 
                 for i, _ in enumerate(as_completed(futures)):
-                    future.result()
-                    logger.debug(f'{i + 1} out of {len(futures)} data handlers'
+                    try:
+                        future.result()
+                    except Exception as e:
+                        logger.error('Error normalizing data handler number '
+                                     f'{futures[future]}')
```
grantbuster, line 488: IMO best practice is to assign your error message to a variable and then do 
```
logger.error(msg) 
raise RuntimeError(msg) from e

File: sup3r/preprocessing/data_handling.py [hunk 5]
```diff
@@ -76,12 +77,199 @@ def get_handler_class(file_paths):
...
             HandlerClass = DataHandlerH5SolarCC
+    elif not is_time_series(file_paths):
+        HandlerClass = DataHandlerNCforCC
     else:
         HandlerClass = DataHandlerNC
     return HandlerClass
 
 
-class DataHandler(FeatureHandler):
+class InputHandler:
```
grantbuster, line 87: I think this is commonly called a [Mixin class](https://www.thedigitalcatonline.com/blog/2020/03/27/mixin-classes-in-python/#:~:text=Mixin%20classes%20in%20Python%C2%B6&text=In%20Python%2C%20so%2Dcalled%20mixins,with%20the%20other%20parent%20classes.), you should refer to it / name it as such

sup3r/sup3r.py
```diff
@@ -94,3 +98,69 @@ def _init_loggers(self, loggers=None, log_file='sup3r.log',
 
             logger.info('Active python environment versions: \n{}'
                         .format(pprint.pformat(VERSION_RECORD, indent=2)))
+
+    @classmethod
+    def init_pass_collect(cls, out_dir, file_paths, model_path, **kwargs):
```
grantbuster, line 103: You should define what these kwargs go to, otherwise the API is difficult to read. You might also want to break this up into foward_pass_kwargs and collect_kwargs and define where the user can find a docstring of these options (consider linking to the API docs). 

Honestly I'm curious if we even need this high level SUP3R class... All it does is make directories and sets up configs? And honestly the user has to do all the work of setting up the configs anyway and just feeding them into the kwargs field so im not convinced this saves a ton of time.

grantbuster (changes requested):

sup3r/sup3r.py
grantbuster: For the nsrdb, the high level NSRDB class is a common entry point for basically the entire software. This is more of a project setup utility... I'd suggest we name it as such and put it in the utilities or config sub modules.

bnb32: Yeah thats what I had in mind when I copied this SUP3R class in here. Should we just remove it after moving this method?

grantbuster: Yeah I dont see a need to have a collection of entry points for all the sub-modules. For the NSRDB it was really to link the pipeline/cli to the pure python class methods

sup3r/pipeline/forward_pass.py
```diff
@@ -100,22 +100,18 @@ def __init__(self, file_paths,
...
+            processes will be serialized.
+        time_chunk_size : int
             Size of chunks to split time dimension into for parallel data
             extraction. If running in serial this can be set to the size
             of the full time index for best performance.
-        cache_file_prefix : str
-            Prefix of path to cached feature data files
+        cache_pattern : str | None
+            Pattern for files for saving feature data. e.g.
+            file_path_{feature}.pkl Each feature will be saved to a file with
```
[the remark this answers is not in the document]
bnb32: That gets a little annoying if we want to write new cache files that dont exist no? That's why I've added a bunch of format keys for cache files in the InputMixIn class.

sup3r/preprocessing/batch_handling.py
grantbuster: actually this should be logger.exception which logs the whole stack trace

grantbuster (approved):

PR #59, opened 2022-07-20
Gb/fwp edits

grantbuster:


sup3r/preprocessing/feature_handling.py
```diff
@@ -1388,31 +1395,46 @@ def lookup(cls, feature, attr_name):
...
                 break
+
         if out is None:
             return None
 
         if not isinstance(out, str):
             return getattr(out, attr_name, None)
+
         elif attr_name == 'inputs':
+
```
bnb32, line 1425: good catch. missed this edge case.

bnb32:
This all looks good. Could we add a test for the issue you ran into in the FeatureHandler? I guess just checking that requesting features without heights/pressure works. The logic in there is getting a little complicated.

grantbuster:
Agreed! Yeah my intent was just that if the variable name is in the handler directly, don't add the height/pressure stuff. I can add another test (makes me so happy you want more tests)

sup3r/pipeline/forward_pass.py
```diff
@@ -918,17 +996,30 @@ def _run_parallel(self, data, max_workers=None):
...
-                    self.hr_crop_slices)):
-                future = exe.submit(
-                    ForwardPass.forward_pass_chunk,
-                    data_chunk=self.data_handler.data[slp],
-                    crop_slices=shc, model_path=self.model_path)
+            zip_iter = zip(self.hr_slices, self.lr_pad_slices,
+                           self.hr_crop_slices)
+            for i, (sh, slp, shc) in enumerate(zip_iter):
+
+                future = exe.submit(ForwardPass.forward_pass_chunk,
```
bnb32, line 1006: I see that you only include the option to pass a loaded model for the serial routine. I experimented with doing this for the parallel routine and got layer errors about mismatched sizes with cached data. I'm not sure the serial routine will avoid this problem either?

bnb32:
Also, is this only going to work for a datetime.datetime or pd.DateIndex object? https://github.com/NREL/sup3r/blob/4e25f19068dbcd94de9a3b867efddc3ca64540fe/sup3r/preprocessing/data_handling.py#L293

sup3r/pipeline/forward_pass.py
grantbuster: I havent been running on cached data but yeah my thought right now is pass the model in for serial and load the model in parallel workers. It's worked for me but I havent been running with cached data. Is there a test for that?

grantbuster:
Yeah is there a case where time index is not a datetimeindex? Docs dont specify but i would always expect that?

bnb32:
> Yeah is there a case where time index is not a datetimeindex? Docs dont specify but i would always expect that?

Yeah the time indices coming from wrf data tend to be np.datetime64. I would just add a check `isinstance(t,np.datetime64)` and convert with `pd.Timestamp`

grantbuster:
okay @bnb32 made some mods, added some tests, should be good to go

bnb32:
> okay @bnb32 made some mods, added some tests, should be good to go

Test looks awesome, thanks!

bnb32 (approved):

PR #61, opened 2022-07-21
Gb/more multi step

grantbuster:
1. Integration of multi step model with forward pass
2. improved solar CC loss function
3. fixed up file handling and collection.


sup3r/pipeline/forward_pass.py
```diff
@@ -867,19 +867,27 @@ def forward_pass_chunk(data_chunk, crop_slices,
...
             model = model_class.load(*model_args, verbose=False)
 
-        data_chunk = np.expand_dims(data_chunk, axis=0)
+        if isinstance(model, sup3r.models.SpatialThenTemporalGan):
+            i_lr_t = 0
+            i_lr_s = 1
+            data_chunk = np.transpose(data_chunk, axes=(2, 0, 1, 3))
+        else:
+            i_lr_t = 3
+            i_lr_s = 1
```
bnb32, line 876: Can you explain what's happening here? No change needed just trying to trace the logic.

tests/test_output_handling.py
```diff
@@ -71,21 +74,59 @@ def test_invert_uv_inplace():
     assert np.allclose(data[..., 1], wd)
 
 
-def test_forward_pass_collection():
-    """Test forward pass collection."""
+def test_h5_out_and_collect():
```
bnb32, line 77: good test!

sup3r/pipeline/forward_pass.py
grantbuster: Yeah so if you have a spatial model first, you want the shape to be (n_obs, spatial, spatial, n_features) where n_obs is actually n_temporal. So this piece of logic sets the index for the coarse temporal / spatial shape so that you can do a good check to make sure the output is enhanced properly.

sup3r/postprocessing/collection.py
```diff
@@ -11,11 +11,10 @@
 
 from rex.utilities.execution import SpawnProcessPool
 from rex.utilities.loggers import init_logger
-from rex import Outputs
 from rex.utilities.fun_utils import get_fun_call_str
 
 from sup3r.pipeline import Status
-from sup3r.postprocessing.file_handling import H5_ATTRS
+from sup3r.postprocessing.file_handling import H5_ATTRS, RexOutputs
```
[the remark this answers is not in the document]
grantbuster: So RexOutputs is just a subclass of rex.handlers.outputs where we update the operative package name / version and we update the "full version record" attribute to reflect additional versions that are important to sup3r. We do this in reV and nsrdb and i think it's good practice. See the details here: 

https://github.com/NREL/sup3r/blob/c4a52b2bf466d93b7e14a115f134bf63bae3bea2/sup3r/postprocessing/file_handling.py#L55-L64

tests/test_output_handling.py
grantbuster: thanks! some stuff broke with my updates so i thought it would be better to make our own files on the fly.

bnb32 (approved):

PR #62, opened 2022-07-22
Bnb/misc

bnb32:


sup3r/models/base.py
```diff
@@ -77,6 +76,8 @@ def __init__(self, gen_layers, disc_layers, loss='MeanSquaredError',
             normalization scheme between transfer learning domains.
         name : str | None
             Optional name for the GAN.
+        verbose : bool
```
grantbuster, line 79: I dont think you need this here. I added these flags to the load() kwargs.

sup3r/pipeline/forward_pass.py
```diff
@@ -755,7 +754,7 @@ def __init__(self, strategy, node_index=0):
             logger.error(msg)
             raise KeyError(msg)
 
-        self.model = model_class.load(*self.model_args)
+        self.model = model_class.load(*self.model_args, verbose=False)
```
grantbuster, line 757: I think the first model load in the non-parallel section should print the model versions (verbose=True)

sup3r/pipeline/forward_pass.py
```diff
@@ -1050,26 +1026,29 @@ def run(self):
...
-            data = self._run_serial(data)
+            self._run_serial()
         else:
-            data = self._run_parallel(data, max_workers)
+            self._run_parallel(max_workers)
 
         logger.info('All forward passes are complete.')
-        data = data[:, :, self.ti_hr_crop_slice, :]
+        self.data = self.data[:, :, self.ti_hr_crop_slice, :]
+        t_chunk = self.strategy.t_enhance * self.strategy.fwp_chunk_shape[-1]
```
grantbuster, line 1041: I think we had a misunderstanding with the output write chunk shape. Chunk size should not be calculated here and passed in. It should only be a function of the dataset name and dtype. See my suggestion in the file handler module.

sup3r/postprocessing/file_handling.py
```diff
@@ -350,7 +355,8 @@ def write_output(cls, data, features, low_res_lat_lon,
                 flat_data = data[..., i].reshape((-1, len(times)))
                 flat_data = np.transpose(flat_data, (1, 0))
                 Outputs.add_dataset(out_file, f, flat_data,
-                                    dtype=attrs['dtype'], attrs=attrs)
+                                    dtype=attrs['dtype'], attrs=attrs,
+                                    chunks=chunks)
```
grantbuster, line 359: okay so i dont think chunks should be an input to this method, but instead be part of the H5_ATTRS lookup so you can set the chunk shape per dataset/dtype. So something like windspeed will always be uint16 with chunk shape (2000, 500) which is 2MB. So right here you would just have `chunks=attrs.get('chunks', DEFAULT_CHUNKS)`

Also this might not be fun but you should rebase on my branch after i merge since i made some important edits to the file handlers.

sup3r/preprocessing/batch_handling.py
```diff
@@ -127,7 +133,8 @@ def get_coarse_batch(cls, high_res,
         if t_enhance != 1:
             low_res = temporal_coarsening(low_res, t_enhance,
                                           temporal_coarsening_method)
-
+        if smoothing is not None:
+            low_res = gaussian_filter(low_res, smoothing, mode='constant')
```
grantbuster, line 137: Is this what you're thinking for making the coarsened data look more like coarse WRF outputs with less high-res turbulence?

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: Yeah good point

sup3r/postprocessing/file_handling.py
grantbuster: Looks like you did exactly this, nevermind!

sup3r/models/base.py
bnb32: Yeah but I don't want the version info logged for every single forward pass chunk

grantbuster: Seeing your most recent commit, I get it now. I think we can actually just move this logging statement into the load() method. There's no reason to print versions on init, just when you're loading. Thoughts?

sup3r/preprocessing/batch_handling.py
bnb32: Yeah exactly

sup3r/models/base.py
bnb32: Yeah I like that

grantbuster (approved):

PR #65, opened 2022-07-26
Bnb/misc

bnb32:
Some small tweaks to accommodate new wtk data handling. Added physical limit enforcement on output handling.


sup3r/pipeline/forward_pass.py
```diff
@@ -783,14 +793,24 @@ def __init__(self, strategy, node_index=0):
...
-        self.data_handler_class = strategy.data_handler_class
+        fwp_out_shape = len(self.output_features)
+        fwp_out_shape *= np.product(self.data_shape)
+        fwp_out_shape *= self.strategy.s_enhance**2 * self.strategy.t_enhance
+        fwp_out_mem = 4 * fwp_out_shape
+        mem = psutil.virtual_memory()
+        msg = (f'Full size of forward pass output ({fwp_out_mem / 1e9} GB) is '
+               'too large to hold in memory. Run with smaller '
+               'fwp_chunk_shape[2] or smaller spatial extent. ')
+        assert (mem.total - mem.used) > fwp_out_mem, msg
```
grantbuster, line 804: Recommend this as a warning instead of an assertion error. I'm pretty sure that sometimes python can hold memory and not free it, but it can use this memory for allocating new data. In any case, shouldnt we get an OOM error pretty quickly?

Also recommend you log the mem available in the msg for comparison purposes.

sup3r/postprocessing/file_handling.py
```diff
@@ -27,7 +27,8 @@
 H5_ATTRS = {'windspeed': {'scale_factor': 100.0,
                           'units': 'm s-1',
                           'dtype': 'uint16',
-                          'chunks': (2000, 500)},
+                          'chunks': (2000, 500),
+                          'min': 0},
             'winddirection': {'scale_factor': 100.0,
```
grantbuster, line 32: min/max on winddirection should be 0, 360 right?

sup3r/postprocessing/file_handling.py
```diff
@@ -27,7 +27,8 @@
 H5_ATTRS = {'windspeed': {'scale_factor': 100.0,
                           'units': 'm s-1',
                           'dtype': 'uint16',
-                          'chunks': (2000, 500)},
+                          'chunks': (2000, 500),
+                          'min': 0},
```
grantbuster, line 31: Probably worth setting a very high max windspeed even if there really isn't one... 120 is what we use for passing ws to SAM.

sup3r/postprocessing/file_handling.py
```diff
@@ -39,7 +40,9 @@
             'relativehumidity': {'scale_factor': 100.0,
                                  'units': 'percent',
                                  'dtype': 'uint16',
-                                 'chunks': (2000, 500)},
+                                 'chunks': (2000, 500),
+                                 'max': 100,
+                                 'min': 0},
             'pressure': {'scale_factor': 0.1,
                          'units': 'Pa',
```
grantbuster, line 47: Can you add physical ranges to all of the variables? Except maybe BVF stuff. With ML stuff it's totally worth protecting against spurious outputs. The ranges can be very generous. For example: 

https://github.com/NREL/rex/blob/5f15a1fc7ee738716425b52fdb87757f34d24881/rex/sam_resource.py#L56-L77

sup3r/postprocessing/file_handling.py
```diff
@@ -84,6 +87,31 @@ class OutputHandler:
...
+        -------
+        data : ndarray
+            Array of feature data with physical limits enforced
+        """
+        for i, f in enumerate(features):
+            attrs = H5_ATTRS[Feature.get_basename(f)]
+            max = attrs.get('max', np.inf)
+            min = attrs.get('min', -np.inf)
+            data[data[..., i] < min, i] = min
+            data[data[..., i] > max, i] = max
```
grantbuster, line 112: I think the `np.maximum`, `np.minimum` operators will be more efficient than creating a mask array

sup3r/preprocessing/feature_handling.py
```diff
@@ -707,97 +704,8 @@ def compute(cls, data, height):
...
 
         """
-        _, v = transform_rotate_wind(
-            data[f'windspeed_{height}m'],
-            data[f'winddirection_{height}m'],
-            data['lat_lon'])
-        return v
-
-
-class UWindNsrdb(DerivedFeature):
```
grantbuster, line 717: why dont we need these anymore? Dont we still need these if we want u/v from NSRDB instead of windspeed/direction? Not sure if we have a test for that.

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: Well currently the array to hold the fwp output is declared after the data extraction so it can take some time to get an OOM error. I could just move the declaration though.

bnb32: Actually I probably don't want to move the declaration bc that will restrict the mem available for the data extraction.

sup3r/preprocessing/feature_handling.py
bnb32: I added 'windspeed' -> 'wind_speed', and 'winddirection' -> 'wind_direction' in the feature registry for the solar data handler, so this enables use of UWind, VWind. I added a test in test_feature_handling also.

sup3r/postprocessing/file_handling.py
bnb32: Do we want to do that though? 540 can be converted to 180 but if we force 540->360 that's no longer possible. There shouldn't be values outside of 0, 360 if the output handler is reached through the pipeline but if we use it directly to write arbitrary data we'd want to preserve the periodicity right?

sup3r/postprocessing/file_handling.py
bnb32: Yeah sure thing

sup3r/postprocessing/file_handling.py
grantbuster: When you convert u/v to windspeed / dir you fix that ceiling/periodicity issue right? I think the output handler should always enforce physical limits. This is is especially important if you set your scale factor close to the bitrange. If you dont enforce a physical ceiling, you could have a bit overflow and end up with garbage.

sup3r/pipeline/forward_pass.py
grantbuster: Okay, well i dont love it but lets see how it goes haha

bnb32: This is really just to check if the data array that holds the fwp output is bigger than the total mem. If we are declaring an 900 GB array on a 800 GB machine itll fail no matter what magic python does

sup3r/postprocessing/file_handling.py
bnb32: Yeah the conversion enforces that range but I was wondering about using the output handler without doing a pipeline run (which then wouldnt include that conversion). But that latter point makes sense.

sup3r/preprocessing/feature_handling.py
bnb32: https://github.com/NREL/sup3r/blob/bcac12a5496611e6376f1cceb3df001b99f237ab/sup3r/preprocessing/data_handling.py#L2095

sup3r/pipeline/forward_pass.py
grantbuster: *Bigger than the _available_ memory. psutil is not perfect and I'd always rather be able to try stuff with a warning instead of the code not letting me.

sup3r/preprocessing/feature_handling.py
bnb32: https://github.com/NREL/sup3r/blob/bcac12a5496611e6376f1cceb3df001b99f237ab/sup3r/preprocessing/data_handling.py#L2095:#L2098

sup3r/postprocessing/file_handling.py
grantbuster: yeah i'd rather protect the physical range rather than allowing an unspecified range. Honestly the rex outputs handler should have protection against bit overflow. I could have sworn i wrote that somewhere at some point but i cant find it.

sup3r/preprocessing/feature_handling.py
grantbuster: ok sweet!

sup3r/pipeline/forward_pass.py
bnb32: Yeah I could change it to total mem instead of available?

grantbuster: I like that!

sup3r/postprocessing/file_handling.py
```diff
@@ -46,7 +51,9 @@
             'pressure': {'scale_factor': 0.1,
                          'units': 'Pa',
                          'dtype': 'uint16',
-                         'chunks': (2000, 500)},
+                         'chunks': (2000, 500),
+                         'min': 0,
+                         'max': 100000},
```
grantbuster, line 56: isnt this just 1 atm?

grantbuster (changes requested):

sup3r/postprocessing/file_handling.py
bnb32: yeah missed a 1. good catch

grantbuster: No what i'm saying is that WRF regularly produces pressures > 1atm so it shouldnt be 101325 Pa (1atm) it should be more like 1.5e5 Pa

bnb32: Yeah I meant I missed a one to make it 1.1e5 but I'll increase it further

grantbuster: Oh gotcha, i saw 1.01e5 and thought you were setting the ceiling at exactly 1atm. sounds good!

sup3r/pipeline/forward_pass.py
```diff
@@ -795,6 +795,12 @@ def __init__(self, strategy, node_index=0):
         self.load_workers = kwargs['load_workers']
         self.output_workers = kwargs['output_workers']
 
+        print('Cropped slices')
+        print('ti_hr_crop_slice')
+        print(self.ti_hr_crop_slice)
```
grantbuster, line 800: @bnb32 did you mean to include these?

bnb32: oops

grantbuster (approved):

PR #68, opened 2022-07-29
added topography raster extraction utility class with tests

grantbuster:


sup3r/preprocessing/data_handling.py
```diff
@@ -623,8 +624,11 @@ def time_chunk_size(self):
         if self._time_chunk_size is None:
             step_mem = self.feature_mem * len(self.raw_features)
             step_mem /= len(self.time_index)
-            self._time_chunk_size = np.min([np.int(1e9 / step_mem),
-                                            self.n_tsteps])
+            if step_mem == 0:
```
bnb32, line 627: this should probably just be n_tsteps, although it won't really matter.

grantbuster: step_mem -> n_tsteps?

grantbuster: or you're saying i should set `self._time_chunk_size = self.n_tsteps`?

bnb32: Yeah time_chunk_size = n_steps. It'll cap to that anyway with 1e9.

bnb32 (approved):
This looks good. I'm wondering you didn't use DataHandlerH5 to extract the source topo data and use hr_spatial_coarsen instead of using the nn aggregation though?

grantbuster:
The sup3r data handler classes spend a lot of time doing rasterization which will fail when sourcing topo data from the full-disc NSRDB which is flattened and sparsified (for example). It makes more sense and is more flexible to use the rex.Resource handler and keep the elevation data flattened.

bnb32:
> The sup3r data handler classes spend a lot of time doing rasterization which will fail when sourcing topo data from the full-disc NSRDB which is flattened and sparsified (for example). It makes more sense and is more flexible to use the rex.Resource handler and keep the elevation data flattened.

Gotcha, makes sense.

PR #69, opened 2022-08-01
Bnb/elev

bnb32:


grantbuster (changes requested):
A question and a request:

1. Is the "topography" feature understood as a train only feature? Do you have to add it to the `DataHandler.TRAIN_ONLY_FEATURES` tuple?
2. Can you add a simple test? Just to make sure it extracts, gets batched, and doesn't get passed to the true high res dataset.

bnb32:
> A question and a request:
> 
>     1. Is the "topography" feature understood as a train only feature? Do you have to add it to the `DataHandler.TRAIN_ONLY_FEATURES` tuple?
> 
>     2. Can you add a simple test? Just to make sure it extracts, gets batched, and doesn't get passed to the true high res dataset.

1. No good catch.
2. Yeah for sure.

grantbuster (approved):

PR #71, opened 2022-08-02
Gb/qa

grantbuster:


sup3r/pipeline/forward_pass.py
```diff
@@ -46,9 +44,9 @@ class ForwardPassStrategy(InputMixIn):
 
     def __init__(self, file_paths, model_args, s_enhance, t_enhance,
                  fwp_chunk_shape, spatial_pad, temporal_pad,
+                 temporal_slice=slice(None),
```
bnb32, line 47: while we're editing this could we add the hr_spatial_coarsen arg to pass to the data handler?

grantbuster: Yeah can do, whats the use case? Would this be for doing validation with data thats already high res?

bnb32: Yeah that exactly.

bnb32 (approved):
Very clean. Looks great.

sup3r/pipeline/forward_pass.py
grantbuster: cool, done!

PR #73, opened 2022-08-05
Bnb/fwp topo

bnb32:
Refactored slicing to accommodate exogenous data. Added simple exo data handler. Added multi step with topo test. Probably still needs some refinement on the exo data handler and accounting for more general multi step models. Also needs s_enhance and t_enhance added to meta write. Open to suggestions on these improvements.

### DIFF
--- a/sup3r/models/base.py
+++ b/sup3r/models/base.py
@@ -411,8 +411,9 @@ class Sup3rGan(AbstractSup3rGan):
         """
         return self.generator.weights
 
+    # pylint: disable=W0613
     def generate(self, low_res, norm_in=True, un_norm_out=True,
-                 exogenous_features=None):
+                 exogenous_data=None):
         """Use the generator model to generate high res data from low res
         input. This is the public generate function.
 
@@ -429,7 +430,7 @@ class Sup3rGan(AbstractSup3rGan):
         un_norm_out : bool
            Flag to un-normalize synthetically generated output data to physical
            units
-        exogenous_features : None
+        exogenous_data : None
            Placeholder to match multistep model generate function signature
 
         Returns
--- a/sup3r/models/base.py
+++ b/sup3r/models/base.py
@@ -442,7 +442,7 @@ class Sup3rGan(AbstractSup3rGan):
             (n_obs, spatial_1, spatial_2, n_features)
             (n_obs, spatial_1, spatial_2, n_temporal, n_features)
         """
-        low_res = (low_res if exogenous_data is None
+        low_res = (low_res if not exogenous_data
                    else np.concatenate((low_res, exogenous_data), axis=-1))
 
         if norm_in and self._means is not None:
--- a/sup3r/models/base.py
+++ b/sup3r/models/base.py
@@ -411,7 +411,6 @@ class Sup3rGan(AbstractSup3rGan):
         """
         return self.generator.weights
 
-    # pylint: disable=W0613
     def generate(self, low_res, norm_in=True, un_norm_out=True,
                  exogenous_data=None):
         """Use the generator model to generate high res data from low res
@@ -430,8 +429,10 @@ class Sup3rGan(AbstractSup3rGan):
         un_norm_out : bool
            Flag to un-normalize synthetically generated output data to physical
            units
-        exogenous_data : None
-           Placeholder to match multistep model generate function signature
+        exogenous_data : ndarray | None
+            Exogenous data array, usually a 4D or 5D array with shape:
+            (n_obs, spatial_1, spatial_2, n_features)
+            (n_obs, spatial_1, spatial_2, n_temporal, n_features)
 
         Returns
         -------
@@ -441,6 +442,8 @@ class Sup3rGan(AbstractSup3rGan):
             (n_obs, spatial_1, spatial_2, n_features)
             (n_obs, spatial_1, spatial_2, n_temporal, n_features)
         """
+        low_res = (low_res if exogenous_data is None
+                   else np.concatenate((low_res, exogenous_data), axis=-1))
 
         if norm_in and self._means is not None:
             low_res = low_res.copy()
--- a/sup3r/models/base.py
+++ b/sup3r/models/base.py
@@ -411,7 +411,8 @@ class Sup3rGan(AbstractSup3rGan):
         """
         return self.generator.weights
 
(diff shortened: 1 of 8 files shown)


sup3r/models/multi_step.py
```diff
@@ -238,7 +253,11 @@ def generate(self, low_res, norm_in=True, un_norm_out=True):
             low_res = self._normalize_input(low_res)
```
grantbuster, line 253: I think the exogenous data for model 0 needs to be attached to the low res data before normalization

sup3r/models/base.py
```diff
@@ -428,6 +430,8 @@ def generate(self, low_res, norm_in=True, un_norm_out=True):
         un_norm_out : bool
            Flag to un-normalize synthetically generated output data to physical
            units
+        exogenous_data : None
+           Placeholder to match multistep model generate function signature
```
grantbuster, line 434: Is the placeholder necessary? Why not just add a statement doing concatenation to the low res array in this method so that in theory you could provide exo data to a single sup3rmodel?

sup3r/pipeline/forward_pass.py
```diff
@@ -31,6 +32,413 @@
 logger = logging.getLogger(__name__)
 
 
+class ForwardPassSlicer:
```
grantbuster, line 35: Nice i like this delegation of slicing responsibility to a single class and the encapsulation with properties is great!

sup3r/pipeline/forward_pass.py
```diff
@@ -803,6 +972,17 @@ def __init__(self, strategy, node_index=0):
...
         self.load_workers = kwargs['load_workers']
         self.output_workers = kwargs['output_workers']
+        self.exo_kwargs = kwargs['exo_kwargs']
+        self.exo_slices = kwargs['exo_slices']
+
+        if self.exo_kwargs is not None:
+            self.exo_features = self.exo_kwargs.get('features', [])
+            self.features = [f for f in self.features
+                             if f not in self.exo_features]
+            self.exo_handler = ExogenousDataHandler(**self.exo_kwargs)
```
grantbuster, line 1007: You probably dont need to assign the handler to an attribute since it's not used elsewhere, but mostly you just want to watch out for conditional attributes - this attribute does not exist at all if exo_kwargs is None

sup3r/preprocessing/exogenous_data_handling.py
```diff
@@ -0,0 +1,83 @@
...
+    enhancement steps."""
+
+    def __init__(self, file_paths, features, source_h5, s_enhancements,
+                 agg_factors, target=None, shape=None, raster_file=None,
+                 max_delta=20, input_handler=None):
+        """
+        Parameters
+        ----------
+        file_paths : str | list
+            A single source h5 wind file to extract raster data from or a list
```
grantbuster, line 24: This isn't always a wind file, and actually you might want a clause saying "this is typically low-res WRF output or GCM netcdf data files that is source low-resolution data intended to be sup3r resolved"... Just to further differentiate it from source_h5. 

And actually I realize now that I wrote this docstring originally for the TopoExtract utility so if you could update it there too that would be great :)

sup3r/preprocessing/exogenous_data_handling.py
```diff
@@ -0,0 +1,83 @@
...
+            data handler class to use for input data. Provide a string name to
+            match a class in data_handling.py. If None the correct handler will
+            be guessed based on file type and time series properties.
+        """
+
+        self.features = features
+        self.s_enhancements = [1] + s_enhancements
+        self.agg_factors = [1] + agg_factors
+        self.data = []
+        for i, (_, _) in enumerate(zip(s_enhancements, agg_factors)):
```
grantbuster, line 67: This seems like a weird way to iterate through these two lists haha

You should probably also add an assert statement to make sure they're of the same length with a clear error message

sup3r/preprocessing/exogenous_data_handling.py
```diff
@@ -0,0 +1,83 @@
...
+            A single source h5 wind file to extract raster data from or a list
+            of netcdf files with identical grid. The string can be a unix-style
+            file path which will be passed through glob.glob
+        features : list
+            List of exogenous features to extract from source_h5
+        source_h5 : str
+            Filepath to source wtk or nsrdb file to get hi-res (2km or 4km)
+            data from which will be mapped to the enhanced grid of
+            the file_paths input
+        s_enhancements : list
```
grantbuster, line 33: So if there's two spatial enhancement steps and a single temporal enhancement step, should s_enhancements be [5, 5, None] or just [5, 5]? Could use clarification on if the length of these args should match the number of models or just the number of spatial steps. I would prefer the former as it is more explicit, but I dont think a None entry is handled below.

grantbuster (changes requested):

sup3r/models/multi_step.py
bnb32: good catch!

sup3r/pipeline/forward_pass.py
bnb32: Thanks!

sup3r/preprocessing/exogenous_data_handling.py
```diff
@@ -0,0 +1,83 @@
...
+            For example, if file_paths has 100km data and s_enhance is 4, this
+            class will output a feature raster corresponding to the file_paths
+            grid enhanced 4x to ~25km
+        agg_factor : list
+            List of factors by which to aggregate the topo_source_h5 elevation
+            data to the resolution of the file_paths input enhanced by
+            s_enhance. For example, if file_paths has 100km data and s_enhance
+            is 4 resulting in a desired resolution of ~25km and topo_source_h5
+            has a resolution of 4km, the agg_factor should be 6 so that 6x 4km
+            cells are averaged to the ~25km enhanced grid.
```
grantbuster, line 45: Hey also can you update this docstring here and in the topo extractor? The agg factor for this example should be 6 in either spatial axis, so the agg_factor should actually be 6**2=36

sup3r/models/base.py
bnb32: Yeah good call. I think my initial approach made this trick but that's been resolved.

grantbuster (approved):

PR #75, opened 2022-08-08
Gb/smooth ignore

grantbuster:


grantbuster:
@bnb32 you should note that the gaussian filter mode has a big effect on the borders of a smoothed sample

sup3r/preprocessing/batch_handling.py
```diff
@@ -95,7 +95,9 @@ def get_coarse_batch(cls, high_res,
                          temporal_coarsening_method='subsample',
                          output_features_ind=None,
                          output_features=None,
-                         smoothing=None):
+                         training_features=None,
```
bnb32, line 98: This is a good idea!

sup3r/preprocessing/batch_handling.py
```diff
@@ -991,47 +1028,57 @@ def __next__(self):
...
+        if self.smoothing is not None:
+            feat_iter = [j for j in range(low_res.shape[-1])
+                         if self.training_features[j]
+                         not in self.smoothing_ignore]
+            for i in range(low_res.shape[0]):
+                for j in feat_iter:
+                    low_res[i, ..., j] = gaussian_filter(low_res[i, ..., j],
+                                                         self.smoothing,
+                                                         mode='nearest')
+
```
bnb32, line 1077: Yeah I should have noted the hard coded method before. I assume nearest sorted some of the boundary issues?

bnb32 (approved):

sup3r/preprocessing/batch_handling.py
grantbuster: Yeah so constant just pads the boundaries with 0 (default) so you get a low-valued border. Nearest does a NN extrapolation so the boundary is seamless.

PR #78, opened 2022-08-10
Bnb/spatial dc

bnb32:


sup3r/models/base.py
```diff
@@ -717,12 +729,74 @@ def meta(self):
...
+        ----------
+        s_enhance : int
+            Spatial enhancement factor
+        t_enhance : int
+            Temporal enhancement factor
+        smoothing : float
+            Smoothing parameter used in gaussian filtering of coarsened data
+        """
+        if self.s_enhance is None:
+            self.meta['s_enhance'] = s_enhance
```
grantbuster, line 764: You could make these three if/elif statements into a for loop. Maybe even have the input be kwargs so you could iterate easily.

sup3r/models/base.py
```diff
@@ -717,12 +729,74 @@ def meta(self):
...
+        Parameters
+        ----------
+        s_enhance : int
+            Spatial enhancement factor
+        t_enhance : int
+            Temporal enhancement factor
+        smoothing : float
+            Smoothing parameter used in gaussian filtering of coarsened data
+        """
+        if self.s_enhance is None:
```
grantbuster, line 763: You could make these three if/elif statements into a for loop. Maybe even have the input be kwargs so you could iterate easily.

grantbuster (changes requested):
Looks good!

sup3r/models/base.py
bnb32: yeah good point

grantbuster (approved):

PR #79, opened 2022-08-15
minor refactor to split out extract/compute serial/parallel methods a…

grantbuster:
…nd to improve docstrings

### DIFF
--- a/sup3r/preprocessing/data_handling.py
+++ b/sup3r/preprocessing/data_handling.py
@@ -465,7 +465,7 @@ class DataHandler(FeatureHandler, InputMixIn):
         self._extract_features = None
         self._noncached_features = None
         self._raw_features = None
-        self._raw_data = None
+        self._raw_data = {}
         self._time_chunks = None
 
         msg = (f'Initializing DataHandler {self.input_file_info}. '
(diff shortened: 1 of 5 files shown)


sup3r/utilities/utilities.py
```diff
@@ -841,11 +841,18 @@ def interp_to_level(var_array, lev_array, levels):
     Parameters
     ----------
     var_array : ndarray
-        Array of variable values
+        Array of variable data, for example u-wind in a 4D array of shape
+        (time, vertical, lat, lon)
     lev_array : ndarray
         Array of height or pressure values corresponding to the wrf source data
+        in the same shape as var_array. If this is height and the requested
```
bnb32, line 848: good addition

sup3r/preprocessing/data_handling.py
```diff
@@ -1367,6 +1345,57 @@ def extract_data(self):
                     f'{dt.now() - now}')
         return self.data
 
+    def run_data_extraction(self):
+        """Run the raw dataset extraction process from disk to raw
+        un-manipulated datasets.
+        """
```
bnb32, line 1351: I think we should still have an `if self.extract_features:` here

bnb32 (approved):

PR #84, opened 2022-08-30
Gb/fwp pad

grantbuster:


bnb32 (commented):
This all seems to make sense, although I pored through all the chunking tweaks fairly quickly. High level idea is just reflection padding on the full data volume and then proceeding with the previous chunking logic (with some tweaks to ensure all chunks are the same size)?

grantbuster:
Yep pretty much! So now the edges of any full fwp domain should have better edge behavior and this will also now enable padding at the temporal boundaries.

bnb32 (approved):

PR #85, opened 2022-09-02
Bnb/fwp spatial dist

bnb32:
Early PR just to check tests so I can refine after the weekend. Needs doc strings edited and a test added for fwp + collection before merge.

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -733,6 +733,7 @@ class ForwardPassStrategy(InputMixIn):
         self.output_workers = output_workers
         self.exo_kwargs = exo_kwargs or {}
         self._cache_pattern = cache_pattern
+        self._input_handler_class = None
         self._input_handler_name = input_handler
         self._spatial_coarsen = spatial_coarsen
         self._max_nodes = max_nodes
@@ -907,9 +908,10 @@ class ForwardPassStrategy(InputMixIn):
         _handler_class
             e.g. DataHandlerNC, DataHandlerH5, etc
         """
-        hclass = get_input_handler_class(self.file_paths,
-                                         self._input_handler_name)
-        return hclass
+        if self._input_handler_class is None:
+            self._input_handler_class = get_input_handler_class(
+                self.file_paths, self._input_handler_name)
+        return self._input_handler_class
 
     def __len__(self):
         """Get the number of nodes that this strategy is distributing to"""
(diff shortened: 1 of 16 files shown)


sup3r/pipeline/forward_pass.py
```diff
@@ -311,6 +308,18 @@ def s_hr_slices(self):
                     self._s_hr_slices.append(hr_slice)
         return self._s_hr_slices
 
+    @property
+    def gids(self):
+        """Get gids for spatial coordinates. Used in final data collection."""
```
grantbuster, line 333: Can you describe what this returns? It's a dictionary with keys=node index and values=arrays of spatial gids per node? Are the spatial gids the same as the gids in the final collected output h5? If not, we need to be very careful and maybe rename this.

sup3r/pipeline/forward_pass.py
```diff
@@ -440,6 +449,16 @@ def get_hr_slices(slices, enhancement, padding, step=None):
             hr_slices.append(slice(start, stop, step))
         return hr_slices
 
+    @property
+    def n_spatial_chunks(self):
+        """Get the number of spatial chunks"""
+        return len(self.hr_crop_slices[0])
```
grantbuster, line 493: So is the default now to distribute all the spatial chunks to individual nodes? Is there any control over this besides the fwp chunk shape?

sup3r/pipeline/forward_pass.py
```diff
@@ -786,23 +798,26 @@ def file_ids(self):
             List of file ids for each output file. Will be used to name output
             files of the form filename_{file_id}.ext
         """
-        if self._file_ids is None:
+        if not self._file_ids:
             n_chunks = len(self.time_index)
             n_chunks /= self.fwp_chunk_shape[2]
             n_chunks = np.int(np.ceil(n_chunks))
             self._file_ids = []
             for i in range(n_chunks):
```
grantbuster, line 844: We should rename n_chunks if this is now only the temporal chunks. Otherwise i would assume n_chunks is the total number of temporal chunks * spatial chunks (which it should be).

grantbuster (changes requested):
So my main concern here is that you don't really have any control over the number of nodes you're running on if you're limited to a very small spatiotemporal fwp chunk... We have all of the spatiotemporal chunks, but it's assumed 1-to-1 with nodes. How much effort would it be to configure many-to-1 (chunks to nodes)? Happy to brainstorm over the phone.

sup3r/pipeline/forward_pass.py
bnb32: Yeah its a dictionary of unique indices for each lat/lon pair for forward pass output. The collected output will have a set of gids equal to the full set.

grantbuster: Yeah just add this to a "returns" docstring. Are the gid's the same order as the flattened h5 output?

bnb32: That's a good point. They should be but maybe this could be related to the seam issue

grantbuster (approved):

PR #86, opened 2022-09-07
Bnb/stitching

bnb32:
Not urgent. Figured I'd open the PR since I've moved on to the spatial chunk distribution.

### DIFF
--- a/sup3r/utilities/stitching.py
+++ b/sup3r/utilities/stitching.py
@@ -89,7 +89,7 @@ class Regridder:
 
 
 def get_files(year, month, input_pattern=IN_FILENAME,
-              output_pattern=OUT_FILENAME):
+              output_pattern=OUT_FILENAME, n_domains=4):
     """Get input files for all domains to stitch together, and output file
     name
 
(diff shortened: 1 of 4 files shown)


sup3r/utilities/stitching.py
```diff
@@ -0,0 +1,420 @@
+"""Utilities for stitching south east asia domains"""
+# -*- coding: utf-8 -*-
+# pylint: disable=all
```
grantbuster, line 3: Is there a good reason not to lint this file? It actually looks like you did a great job with formatting and whatnot.

grantbuster (changes requested):
Looks great, minor edits. My main concern is that someone would have no idea where to start in the stitching.py file... Would you want to add an enumerated description of the process in the file docstring? Even just saying which functions to interact with would be helpful.

sup3r/utilities/stitching.py
bnb32: I did this so I didn't have to add xesmf to setup but your suggestion to move to init is a good idea

grantbuster (approved):

PR #87, opened 2022-09-08
Gb/solar dh

grantbuster:


sup3r/models/base.py
```diff
@@ -350,8 +350,8 @@ def set_model_params(self, **kwargs):
                 msg = ('GAN was previously trained with {var}={} but '
                        'received new {var}={}'
                        .format(val, kwargs[var], var=var))
-                logger.error(msg)
-                raise KeyError(msg)
+                logger.warning(msg)
```
bnb32, line 353: yeah good call. had to edit the json for some things weve updated

sup3r/preprocessing/data_handling.py
```diff
@@ -1814,6 +1858,112 @@ def raw_time_index(self):
...
+
+        super().run_data_extraction()
+
+        if get_clearsky:
+            cs_ghi = self.get_clearsky_ghi()
+            for it, tslice in enumerate(self.time_chunks):
+                self._raw_data[it]['clearsky_ghi'] = cs_ghi[..., tslice]
+            self._raw_features.append('clearsky_ghi')
+
+    def get_clearsky_ghi(self):
```
bnb32, line 1881: Should this be put with the other exogenous data handling stuff?

sup3r/preprocessing/data_handling.py
```diff
@@ -2186,6 +2358,66 @@ def feature_registry(cls):
...
+            total_cs_ghi = np.nansum(self.data[:, :, t_slice, i_cs], axis=2)
+            avg_cs_ratio = total_ghi / total_cs_ghi
+            self.daily_data[:, :, d, i_ratio] = avg_cs_ratio
+
+        # remove ghi and clearsky ghi from feature set. These shouldn't be used
+        # downstream for solar cc and keeping them confuses the batch handler
+        logger.info('Finished calculating daily average clearsky_ratio, '
+                    'removing ghi and clearsky_ghi from the '
+                    'DataHandlerH5SolarCC feature list.')
+        ifeats = np.array([i for i in range(len(self.features))
```
bnb32, line 2411: maybe we could add something like COMPUTE_ONLY_FEATURES = (...) and make this removal routine more general. could be placed in the super class also.

sup3r/preprocessing/data_handling.py
```diff
@@ -2166,6 +2314,30 @@ class DataHandlerH5SolarCC(DataHandlerH5WindCC):
...
+            Same positional args as DataHandlerH5
+        **kwargs : dict
+            Same keyword args as DataHandlerH5
+        """
+
+        args = copy.deepcopy(args)  # safe copy for manipulation
+        required = ['ghi', 'clearsky_ghi', 'clearsky_ratio']
+        missing = [dset for dset in required if dset not in args[1]]
+        if any(missing):
+            msg = ('Cannot initialize DataHandlerH5SolarCC without required '
```
bnb32, line 2331: Can you explain this message? The sum stuff, in the context of computing an average, is confusing to me

sup3r/models/base.py
grantbuster: yeah totally. we should give ourselves the power to manipulate these models in crazy ways muahhaha

sup3r/preprocessing/data_handling.py
grantbuster: you know, i thought about that, but the weirdness is that we want the data handler features to always map to the same set of training features that the model expects. So I thought this was a cleaner solution so that once the data handler is initialized it does indeed have the training and output features that one would expect.

bnb32: ok that's fair

sup3r/preprocessing/data_handling.py
grantbuster: So these features are extra weird because they're really required to compute the daily average data, not just for computing the high-res data. So i could generalize this to the top level CC class, but it really does seem like a one-off at this point.

sup3r/preprocessing/data_handling.py
grantbuster: Yeah so the idea is that you cannot actually calculate a daily average clearsky ratio from the timeseries of clearsky ratios (weird). The correct way to calculate a daily average clearsky ratio is by taking the sum (or average) of the daily GHI and dividing by the sum (or average) of the daily clearsky GHI. Does that make sense?

bnb32: Oh yeah, that is weird. My confusion stemmed from thinking you could do the former.

sup3r/preprocessing/data_handling.py
bnb32: Ah yeah, ok makes sense.

bnb32 (approved):

PR #89, opened 2022-09-13
Gb/solar 3step model

grantbuster:


bnb32 (approved):
Surprisingly straightforward review, considering your description in our convo ;)

PR #93, opened 2022-09-16
added simple feature to allow for incremental progress if fwp chunks …

grantbuster:
…fail

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -542,6 +542,7 @@ class ForwardPassStrategy(InputMixIn):
                  overwrite_cache=False,
                  input_handler=None,
                  input_handler_kwargs=None,
+                 incremental=True,
                  max_workers=None,
                  extract_workers=None,
                  compute_workers=None,
@@ -641,6 +642,10 @@ class ForwardPassStrategy(InputMixIn):
             Optional kwargs for initializing the input_handler class. For
             example, this could be {'hr_spatial_coarsen': 2} if you wanted to
             artificially coarsen the input data for testing.
+        incremental : bool
+            Allow the forward pass iteration to skip spatiotemporal chunks that
+            already have an output file (True, default) or iterate through all
+            chunks and overwrite any pre-existing outputs (False).
         max_workers : int | None
             Providing a value for max workers will be used to set the value of
             extract_workers, compute_workers, output_workers, and load_workers.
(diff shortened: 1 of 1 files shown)


sup3r/pipeline/forward_pass.py
```diff
@@ -1513,12 +1513,18 @@ def _run_single_fwd_pass(self):
...
 
     @classmethod
-    def run(cls, strategy, node_index):
+    def run(cls, strategy, node_index, incremental=True):
         """This routine runs forward passes on all spatiotemporal chunks for
         the given node index"""
         for chunk_index in strategy.node_chunks[node_index]:
-            fwp = cls(strategy, chunk_index, node_index)
-            fwp.run_chunk()
+            out_file = strategy.out_files[chunk_index]
```
bnb32, line 1520: Just thought about adding this this morning! Good call

bnb32: Lets add this arg to fwp cli, just in case we want to overwrite easily

grantbuster: yeah thanks, i've been having a lot of random job failures, i think due to concurrent reads? but it's hard to tell whats going on. I think generally lustre is struggling with too much parallel io.... This should help but reducing our reliance on cache files will also be good.

bnb32: Yeah I've had a few as well, since I moved my env to lustre (hmmm). Basically the only overlapping io is to the conda env right? There's no overlapping cache read/write between fwp calls (except for the time index files).

bnb32 (approved):
Just the option to provide increment=False through the config. All good otherwise!

sup3r/pipeline/forward_pass.py
grantbuster: yeah i think in my case i had just had a few random job failures when writing cache files, possibly due to writing too many small cache files to a single OST in parallel. Removing the cache pattern input and clearing the cache dir fixed my jobs.

bnb32: Sweet! whats OST btw?

grantbuster: object storage target (i think). when you stripe a directory it distributes to many OST (up to 30ish on eagle?). single OSTs can get overloaded by parallel io i think.

bnb32: ah gotcha

PR #94, opened 2022-09-20
Bnb/stats

bnb32:
Added collection masking with test. Includes wind stats module base on qa class.


sup3r/postprocessing/collection.py
```diff
@@ -213,7 +213,8 @@ def get_data(self, file_path, feature, time_index, meta, scale_factor,
                 logger.error(e)
                 raise KeyError(e)
 
-            f_data = f[feature][...]
+            mask = f_meta['gid'].isin(meta['gid'])
+            f_data = f[feature][:, mask]
```
grantbuster, line 217: Is this the strategy for doing the sparse collection? We should talk about this... The sparse final meta will likely have a gid column that is enumerated (reset) and is not sparse.

sup3r/postprocessing/collection.py
```diff
@@ -229,7 +230,70 @@ def get_data(self, file_path, feature, time_index, meta, scale_factor,
...
+        with ThreadPoolExecutor(max_workers=max_workers) as exe:
+            for i, fn in enumerate(file_paths):
+                future = exe.submit(cls._get_file_attrs, fn)
+                futures[future] = i
+
+            interval = np.int(np.ceil(len(futures) / 10))
+            for i, future in enumerate(as_completed(futures)):
+                if interval > 0 and i % interval == 0:
+                    mem = psutil.virtual_memory()
+                    logger.info('Collection futures completed: '
```
grantbuster, line 278: This should not be called a collection future. Will be confused with the actual collection futures.

sup3r/postprocessing/collection.py
```diff
@@ -485,6 +592,14 @@ def collect(cls, file_paths, out_file, features, n_writes=1,
...
         job_name : str
             Job name for status file if running from pipeline.
+        join_time : bool
+            Option to split full file list into chunks with each chunk having
+            the same temporal_chunk_index. The number of temporal chunk indices
+            will then be used as the number of writes. Assumes file_paths have
+            a suffix format _{temporal_chunk_index}_{spatial_chunk_index}.h5
+        mask_file : str
+            Path to meta mask containing coordinates to keep from the full
+            file list collected meta
```
grantbuster, line 602: We should talk about this. It looks like you also have a requirement that there's a sparsified "gid" column, which i do not think will be the case. Best way to do this would be to run a KDTree with a very strict distance threshold to determine the mapping.

sup3r/utilities/test_utils.py
```diff
@@ -182,6 +183,29 @@ def tke_spectrum(u, v):
     return E_k
 
 
+def tke_series(u, v):
+    """Longitudinal Turbulent Kinetic Energy Spectrum time series. Gives the
```
grantbuster, line 187: longitudinal? Is that true?

sup3r/utilities/test_utils.py
```diff
@@ -275,3 +299,85 @@ def ramp_rate_dist(u, v, bins=50, range=None):
...
+        alpha = (t / t_enhance - t0) / (t1 - t0)
+        if t1 == low.shape[-1]:
+            tmp = low[..., t0]
+        else:
+            tmp = (1 - alpha) * low[..., t0] + alpha * low[..., t1]
+        high[..., t] = tmp
+    return high
+
+
+def st_interp(low, s_enhance, t_enhance):
```
grantbuster, line 364: we should move all of these things to /sup3r/qa/utilities.py and rename this file pytest_utils.py

sup3r/qa/stats.py
```diff
@@ -0,0 +1,494 @@
...
+            (lat, lon, temporal)
+
+        Returns
+        -------
+        stats : dict
+            Dictionary of stats for wind fields
+        """
+        v_dist = velocity_gradient_dist(u)
+        vort_dist = vorticity_dist(u, v)
+        rr_dist = ramp_rate_dist(u, v)
```
grantbuster, line 432: Hey i actually have some questions about the ramp rate dist function. 

- if you want to keep this as u/v ramps, we should rename it to ws_ramp_rate_dist. 
- We should set the diff screening criteria to a kwarg (`diffs = diffs[(np.abs(diffs) < 10)]`)
- We should set the difference window to a kwarg (6 and 12 hour ramp rates are just as important for power systems work but may not be useful for wind-actuals validation)
- Why do you perform this RMSE operation: `diffs = diffs / np.sqrt(np.mean(diffs**2))`? This makes the output units not values of m/s right?
- I think we should be normalizing the counts output to `counts.sum()` not `counts.max()`. The former is more typical of PDF's isnt it?

sup3r/qa/stats.py
```diff
@@ -0,0 +1,494 @@
...
+        uidx, vidx = self.feature_indices(height)
+        u_lr = self.source_handler.data[..., self.lr_t_slice, uidx]
+        v_lr = self.source_handler.data[..., self.lr_t_slice, vidx]
+        u_itp = st_interp(u_lr, self.s_enhance, self.t_enhance)
+        v_itp = st_interp(v_lr, self.s_enhance, self.t_enhance)
+
+        low_res = self.get_wind_stats(u_lr, v_lr)
+        high_res = self.get_wind_stats(u_hr, v_hr)
+        interp = self.get_wind_stats(u_itp, v_itp)
+        return low_res, high_res, interp
```
grantbuster, line 466: Nice work, this is nice and clean and i can see why this is going to be super useful for all the wind actuals runs.

sup3r/qa/stats.py
```diff
@@ -0,0 +1,494 @@
...
+
+            stats[f'lr_{height}m'] = lr_stats
+            stats[f'hr_{height}m'] = hr_stats
+            stats[f'interp_{height}m'] = interp
+
+        if self.qa_fp is not None:
+            self.export(self.qa_fp, stats)
+        logger.info('Finished Sup3rWindStats run method.')
+
+        return stats
```
grantbuster, line 494: We should probably test our stats module... it's not really QA if we don't is it? Unit tests on the qa utilities would be best.

grantbuster (changes requested):

sup3r/postprocessing/collection.py
bnb32: The initial collection on the entire file list does a masking to sparsify the gids though

sup3r/postprocessing/collection.py
bnb32: Yeah that's a fair point. My talk with Evan seemed to indicate that he'd provide a meta file with gids but maybe not.

sup3r/qa/stats.py
bnb32: Yeah good points about kwargs. The RSME operation is used in the initial superres paper. Also good point about norming by the sum.

sup3r/qa/stats.py
bnb32: Yeah I was waiting for the final collection on wtk-led to finish. I had to rerun due to some of those same issues I showed you yesterday, with the empty chunks. I think I traced this back to n_writes splitting the file list such that the spatial domains for each file list chunk didnt match.

sup3r/qa/stats.py
bnb32: Thanks!

sup3r/utilities/test_utils.py
bnb32: Yeah, it's being averaged over the latitude dimension

sup3r/postprocessing/collection.py
grantbuster: Those gid's will be enumerated 1:N though where N is the final count, or at least that's going to be our desired final output. There's no reason the final meta data should have a record of the original placement in the raster. 

Also, is this mask_file intended to be the final meta data? If so, you should call it that (or "target meta"), and make sure that we are using the full dataframe as the final collected file output (e.g. make sure we're passing through the actual meta data and not just the lat/lon from the fwp).

sup3r/utilities/test_utils.py
grantbuster: ah i see that, interesting!

sup3r/postprocessing/collection.py
bnb32: I provided him meta data with the original gids so I got the feeling that he would return a subset of these as the mask. Either way though I swapped this out for the KDTree approach since that is much more robust. I was thinking of the mask file as not necessarily the final meta, just a set of allowed coordinates - the final meta could still end up being a subset of the allowed coordinates for a general use case.

sup3r/qa/stats.py
grantbuster: That's fine if we want normalized values for the ramp rate / velocity gradient but we should not represent our outputs as windspeeds in m/s then and we should change the output docstring (the outputs are not delta windspeed or delta u but a normalized version of that). The normalization makes sense for problems where you're changing the spatiotemporal resolution but makes the outputs more difficult to interpret when you're benchmarking against a similar dataset with the same target resolution.

sup3r/postprocessing/collection.py
grantbuster: I agree the KDTree will be more robust. I think we should take advantage of this "target_final_meta" to add in the actual meta data provided by evan. Otherwise we would have to add another step to overwrite the meta. Easier to offer that as an option here.

bnb32: I was thinking I'd just add a step to overwrite at the end, since I rely on the original gids to collect the spatial chunks.

bnb32: I suppose I could just replace the gid indexing with the kdtree though.

grantbuster: Yeah you should be using the hi-res raster meta data, build a KDTree from that, and then query using the final meta data. The output array of index values will give you the gids from the high-res raster dataset that need to be collected along with where they belong in the output.

sup3r/postprocessing/collection.py
```diff
@@ -170,6 +173,19 @@ def get_slices(final_time_index, final_meta, new_time_index, new_meta):
...
+    def get_coordinate_indices(cls, target_meta, full_meta):
+        """Get coordindate indices in meta data for given targets"""
+        ll2 = np.vstack((full_meta.latitude.values,
+                         full_meta.longitude.values)).T
+        tree = KDTree(ll2)
+
+        targets = np.vstack((target_meta.latitude.values,
+                             target_meta.longitude.values)).T
+        _, indices = tree.query(targets, distance_upper_bound=0.001)
+        indices = [i for i in indices if i < len(full_meta)]
```
grantbuster, line 186: Just curious, whats the reason for this filtering on max value? The tree query shouldn't return anything greater than the full meta?

grantbuster (approved):

sup3r/postprocessing/collection.py
bnb32: When I was testing this the query returned an index equal to the length of the meta if the query wasn't found. Maybe it only does this when you query a single point?

sup3r/qa/stats.py
bnb32: Yeah good point. I'll update the doc string.

sup3r/postprocessing/collection.py
grantbuster: oh yeah "Missing neighbors are indicated with self.n", i thought it used to be negative but i guess they updated it. Honestly i usually just do a mask with the returned distance and set the bad values to NaN or something else more explicit haha.

bnb32: oh thats a good idea

PR #95, opened 2022-09-30
Bnb/dev

bnb32:
Added stats cli and got process pool for passes on single nodes working.


sup3r/pipeline/forward_pass.py
```diff
@@ -633,6 +636,8 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
         overwrite_cache : bool
             Whether to overwrite cache files storing the computed/extracted
             feature data
+        overwrite_ti_cache : bool
```
grantbuster, line 639: Why wouldnt this just use the `overwrite_cache` arg? Is there any confusing and/or logic here?

sup3r/pipeline/forward_pass.py
```diff
@@ -848,8 +862,7 @@ def get_time_index(self, file_paths, max_workers=None):
         return self.input_handler_class.get_time_index(file_paths,
                                                        max_workers=max_workers)
 
-    @property
-    def single_time_step_files(self):
+    def is_single_ts_files(self):
```
grantbuster, line 865: good method naming

sup3r/pipeline/forward_pass.py
```diff
@@ -1587,42 +1605,54 @@ def run(cls, strategy, node_index):
...
+                            f'{len(strategy.node_chunks[node_index])} '
+                            'complete. Current memory usage is '
+                            f'{mem.used / 1e9:.3f} GB out of '
+                            f'{mem.total / 1e9:.3f} GB total.')
         else:
             logger.debug(f'Running forward passes on node {node_index} in '
                          'parallel with pass_workers='
                          f'{strategy.pass_workers}.')
-            with ThreadPoolExecutor(max_workers=strategy.pass_workers) as exe:
+            with SpawnProcessPool(max_workers=strategy.pass_workers) as exe:
```
grantbuster, line 1626: what kind of compute vs. memory tradeoffs are you seeing here? When the process pool parallelization was after the IO i was seeing no improvement in compute but a penalty in memory (crazy and i cant explain it but it sucked).

sup3r/postprocessing/collection.py
```diff
@@ -466,47 +466,40 @@ def _ensure_dset_in_output(out_file, dset, data=None):
...
+    def _collect_flist(self, feature, masked_meta, time_index, shape,
+                       file_paths, out_file, target_final_meta,
+                       masked_target_meta, max_workers=None):
+        """Collect a dataset from a file list without getting attributes first.
 
         Parameters
         ----------
+        feature : str
+            Dataset name to collect.
+        masked_meta : pd.DataFrame
```
grantbuster, line 478: This method docstring is confusing. What is the difference between `masked_meta` and `masked_target_meta`? Is the file paths a subset (chunk) of the full list of file paths? Is `target_final_meta_file` really a file path? (also target_final_meta_file is misnamed).

sup3r/postprocessing/collection.py
```diff
@@ -578,6 +571,51 @@ def collect_flist(self, file_paths, out_file, feature, sort=False,
...
+
+        Collects data that can be chunked in both space and time.
+
+        Parameters
+        ----------
+        file_paths : list | str
+            Explicit list of str file paths that will be sorted and collected
+            or a single string with unix-style /search/patt*ern.h5. Files
+            should have non-overlapping time_index dataset and fully
+            overlapping meta dataset.
```
grantbuster, line 587: Doesnt need fully overlapping meta anymore, right?

sup3r/postprocessing/collection.py
```diff
@@ -578,6 +571,51 @@ def collect_flist(self, file_paths, out_file, feature, sort=False,
...
+        collect_dir : str
+            Directory of chunked files (flist).
+        out_file : str
+            File path of final output file.
+        dset : str
```
grantbuster, line 592: dset isn't in the arg list

sup3r/postprocessing/collection.py
```diff
@@ -578,6 +571,51 @@ def collect_flist(self, file_paths, out_file, feature, sort=False,
...
+            if out_file does not exist).
+        max_workers : int | None
+            Number of workers to use in parallel. 1 runs serial,
+            None uses all available.
+        target_final_meta_file : str
+            Path to target final meta containing coordinates to keep from the
+            full file list collected meta
+        masked_target_meta : pd.DataFrame
+            Collected meta data with mask applied from target final meta so
+            original gids are preserved.
```
grantbuster, line 607: its weird that you need both of these meta arguments here. Why is that? Is it just to improve performance somehow?

Honestly you should just remove the public `collect_flist()`, it's only called in a single place and only modifies the protected `_collect_flist()` with a single additional function call that could be easily moved to the place `collect_flist()` is called. I'd recommend only keeping the protected method because the arguments are confusing and I think are meant to be called indirectly.

sup3r/preprocessing/data_handling.py
```diff
@@ -437,12 +437,12 @@ class DataHandler(FeatureHandler, InputMixIn):
...
-                 sample_shape=(10, 10, 1), raster_file=None,
+                 sample_shape=(10, 10, 1), raster_file=None, raster_index=None,
                  shuffle_time=False, time_chunk_size=None, cache_pattern=None,
-                 overwrite_cache=False, load_cached=False,
-                 train_only_features=None, max_workers=None,
+                 overwrite_cache=False, overwrite_ti_cache=False,
+                 load_cached=False, train_only_features=None, max_workers=None,
                  extract_workers=None, compute_workers=None, load_workers=None,
-                 norm_workers=None, ti_workers=None):
+                 norm_workers=None, ti_workers=None, handle_features=None):
```
grantbuster, line 445: docstring for handle_features? why provide separately?

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: There are lots of times I dont want to overwrite the time index but do for the other cache.

sup3r/pipeline/forward_pass.py
bnb32: I saw around a 2x speed up with workers=10 vs workers=1 for 100 chunks with (50,50,10) shape.

sup3r/preprocessing/data_handling.py
bnb32: This was to enable precomputing for the forward pass pipeline. When I was getting threads working these repeated calls to xarray caused issues.

sup3r/postprocessing/collection.py
bnb32: Yeah it's to improve performance, since we only have to mask the full meta once. Good suggestion on removing this though

sup3r/postprocessing/collection.py
bnb32: good catch!

sup3r/postprocessing/collection.py
bnb32: Masked meta is for the file list chunk and masked target meta is for the entire file list. Good catch on the misnaming, here it's not a file.

grantbuster (approved):

PR #97, opened 2022-10-04
Gb/wind cc

grantbuster:
@bnb32 one more feature before i want to do a version release. This is a more simple one, just the new windCC model


bnb32 (approved):
Looks good!

PR #98, opened 2022-10-05
Bnb/dev

bnb32:
@grantbuster wanted to include a simple stats module test before the new release


sup3r/pipeline/forward_pass.py
```diff
@@ -667,38 +620,24 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
...
-            pass
+            Dictionary of args to pass to :class:
+            `sup3r.preprocessing.exogeneous_data_handling.ExogenousDataHandler`
+            for extracting exogenous features such as topography for future
+            multistep foward pass
         bias_correct_method : str | None
             Optional bias correction function name that can be imported from
-            the sup3r.bias.bias_transforms module. This will transform the
-            source data according to some predefined bias correction
+            the :meth:`sup3r.bias.bias_transforms` module. This will transform
```
grantbuster, line 636: as much as i love `meth`, i think this should be `mod`

sup3r/pipeline/forward_pass.py
```diff
@@ -794,6 +722,34 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
...
+                                                        False)
+        self.overwrite_ti_cache = input_handler_kwargs.get(
+            'overwrite_ti_cache', False)
+        self.extract_workers = input_handler_kwargs.get('extract_workers',
+                                                        None)
+        self.compute_workers = input_handler_kwargs.get('compute_workers',
+                                                        None)
+        self.load_workers = input_handler_kwargs.get('load_workers', None)
+        self.ti_workers = input_handler_kwargs.get('ti_workers', None)
+        self._cache_pattern = input_handler_kwargs.get('cache_pattern', None)
```
grantbuster, line 751: There's an unofficial pep8 standard that says all attributes should be instantiated in the `__init__`. I've found that doing this makes it a lot easier to track attributes and figure out what they are / where they're defined or modified. Thoughts on moving all of these definitions back up, or making them properties that just retrieve the desired attr from `self.input_handler_kwargs` dynamically?

sup3r/pipeline/pipeline.py
```diff
@@ -48,80 +44,3 @@ def __init__(self, pipeline, monitor=True, verbose=False):
         if 'logging' in self._config:
             init_logger('sup3r.pipeline', **self._config.logging)
             init_logger('reV.pipeline', **self._config.logging)
-
-    @classmethod
-    def init_pass_collect(cls, out_dir, file_paths, model_path,
-                          fwp_kwargs=None, dc_kwargs=None):
-        """Generate config files for forward pass and collection
```
grantbuster, line 55: Never used? Auto-making configs is often redundant because you need to organize the inputs to make the configs in the first place haha.

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: Yeah i got the feeling this was not standard. Will move back up to  init.

grantbuster: Yeah the main thing that sucks if the attrs arent initialized in `__init__` is when you see an attr somewhere random and you cant figure out where it was actually set.

sup3r/postprocessing/collection.py
```diff
@@ -488,12 +490,16 @@ def _collect_flist(self, feature, masked_meta, time_index, shape,
...
+            collected meta. This can be but is not necessarily a subset of the
+            full list of coordinates for all files in the file list. This is
+            used to remove coordinates from the full file list which are not
+            present in the target_final_meta.
         masked_target_meta : pd.DataFrame
-            Collected meta data with mask applied from target_final_meta so
-            original gids are preserved.
+            Dataframe containing the same coordinates contrained in the
+            target_final_meta but with the same gids that are present in the
+            full list of coordinates present in the full file list.
```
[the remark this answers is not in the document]
bnb32: Lol yeah it is confusing. the masked_meta is the same as masked_target_meta but for the subset of files provided. masked_target_meta can be derived from the unmasked meta for the full file list + the final target meta (its just the subset of the full meta with coordinates present in the final target).The only reason target_final_meta is passed here is because of the initialization check for the output file. Which could be removed.

sup3r/pipeline/forward_pass.py
bnb32: Yeah definitely makes sense. Were doing this in ForwardPass.get_chunk_kwargs also though. I wanted to move a lot of that to properties so I suppose now is the time.

grantbuster: Yeah totally your call if you move them to properties or just attributes set in `__init__`. If you move them to properties just make sure to say "retrieved from `input_handler_kwargs` with default of ..." in the docstring

sup3r/pipeline/pipeline.py
bnb32: Haha yeah forgot I had even added this and never used it

grantbuster (approved):

PR #100, opened 2022-10-12
Gb/hr exo

grantbuster:


sup3r/models/multi_step.py
```diff
@@ -452,16 +355,14 @@ def generate(self, low_res, norm_in=True, un_norm_out=True,
         """
         logger.debug('Data input to the 1st step spatial-only '
                      'enhancement has shape {}'.format(low_res.shape))
-        s_exogenous = None
         t_exogenous = None
         if exogenous_data is not None:
-            s_exogenous = exogenous_data[:len(self.spatial_models)]
```
bnb32, line 458: maybe this is still worth keeping just for readability?

bnb32 (approved):
Looks like it was pretty simple to extend the forward pass. Good work.

sup3r/models/multi_step.py
grantbuster: Im trying to avoid the variable-to-variable pointer here: `s_exogenous = exogenous_data` which always seems kind of silly. Is it worth the readability? I could go either way.

grantbuster:
Thanks! yeah the key was realizing the fwp can handle multiple exo layers up to a pretty high resolution and then the multi step model can decide on how to distribute those layers

sup3r/models/multi_step.py
bnb32: Oh yeah thats fair. I could go either way also. Could just pass exo_data[:len(self.spatial_models)] directly into generate.

grantbuster: Yeah but the reason im not doing that now is so that the spatial models can have access to ALL of the exo_data including the 4km data that is technically assigned to the temporal model.

bnb32: Oh yeah I suppose that spatial models need up to `len(self.spatial_models) + 1` if theyre using the concat or add layers.

PR #101, opened 2022-10-12
Bnb/dev

bnb32:
added non surface features for DataHandlerNCforCC

### DIFF
--- a/sup3r/utilities/utilities.py
+++ b/sup3r/utilities/utilities.py
@@ -964,9 +964,16 @@ def interp_var_to_height(data, var, raster_index, heights,
     if len(data[var].dims) == 5:
         raster_index = [0] + raster_index
     logger.debug(f'Interpolating {var} to heights (meters): {heights}')
-    return interp_to_level(unstagger_var(data, var, raster_index, time_slice),
-                           calc_height(data, raster_index, time_slice),
-                           heights)[0]
+    hgt = calc_height(data, raster_index, time_slice)
+    if data[var].dims == ('plev',):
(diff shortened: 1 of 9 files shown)


sup3r/utilities/utilities.py
```diff
@@ -963,10 +963,17 @@ def interp_var_to_height(data, var, raster_index, heights,
...
-                           calc_height(data, raster_index, time_slice),
-                           heights)[0]
+    logger.debug(f'Interpolating {var} to heights (meters): {heights}')
+    hgt = calc_height(data, raster_index, time_slice)
+    if data[var].dims == ('plev',):
+        arr = np.array(data[var])
+        arr = np.expand_dims(arr, axis=(0, 2, 3))
+        arr = np.repeat(arr, hgt.shape[0], axis=0)
+        arr = np.repeat(arr, hgt.shape[2], axis=2)
+        arr = np.repeat(arr, hgt.shape[3], axis=3)
```
grantbuster, line 973: is this where the pressure level magic happens? Looks great, this is exactly why i reached out for help :)

grantbuster (changes requested):
Looks good but I think you still need to add a clause in `FeatureHandler.lookup()` that prioritizes an exact match from the feature registry over the regex match.

bnb32:
> Looks good but I think you still need to add a clause in `FeatureHandler.lookup()` that prioritizes an exact match from the feature registry over the regex match.

This can be accomplished by the ordering in the feature registry but I also added an explicit loop for this.

grantbuster (approved):

PR #103, opened 2022-10-19
Gb/surf min max

grantbuster:


tests/test_data_handling_h5_cc.py
```diff
@@ -91,6 +96,54 @@ def test_solar_handler(plot=False):
...
+        handler = DataHandlerH5SolarCC(res_fp, features_s,
+                                       target=TARGET_S, shape=SHAPE,
+                                       temporal_slice=slice(None, None, 2),
+                                       time_roll=-7,
+                                       val_split=0.1,
+                                       sample_shape=(20, 20, 24),
+                                       max_workers=1)
+
+        assert handler.data.shape[2] % 24 == 0
+        assert handler.val_data.shape[2] % 24 == 0
```
bnb32, line 126: I dont think this will pass. I changed handler.val_data is be None when val_split = 0.0, and val_split is forced to be 0.0 for this data handler, right?

sup3r/preprocessing/feature_handling.py
```diff
@@ -896,6 +896,84 @@ def compute(cls, data, height):
         return data['tas'] - 273.15
 
 
+class TasMin(DerivedFeature):
```
bnb32, line 899: When I see all these tas features using the same celsius conversion I wonder if there is way to combine them. But maybe its better for readability to keep them seperate? Whats your philosophy on this?

tests/test_data_handling_h5_cc.py
grantbuster: Gotcha yeah, I think i must have missed this during our concurrent dev. fixed.

sup3r/preprocessing/feature_handling.py
grantbuster: No philosophy just didn't think about it, good call though, combined with the help of a class attribute

bnb32 (commented):
All looks good

bnb32 (approved):
Combined temp features looks nice

PR #104, opened 2022-10-19
I had added this shift to make sure the lons were monotonically incre…

bnb32:
…asing, which fails when lon crosses 0. Not sure why I thought -180 to 180 range wouldnt be monotonically increasing.


grantbuster (approved):
nice easy fix!

PR #105, opened 2022-10-19
Gb/linear model

grantbuster:


bnb32 (approved):
Nice clean interpolation model.

PR #108, opened 2022-11-07
Bnb/dev

bnb32:
Misc cleaning. Discussed changes to qa utilities.

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -596,47 +596,10 @@ class ForwardPassStrategy(InputMixIn):
             passes for subsequent temporal stitching. This overlap will pad
             both sides of the fwp_chunk_shape. Note that the first and last
             chunks in the temporal dimension will not be padded.
-        temporal_slice : slice | tuple | list
-            Slice defining size of full temporal domain. e.g. If we have 5
-            files each with 5 time steps then temporal_slice = slice(None) will
-            select all 25 time steps. This can also be a tuple / list with
-            length 3 that will be interpreted as slice(*temporal_slice)
         model_class : str
             Name of the sup3r model class for the GAN model to load. The
             default is the basic spatial / spatiotemporal Sup3rGan model. This
             will be loaded from sup3r.models
-        target : tuple
-            (lat, lon) lower left corner of raster. You should provide
-            target+shape or raster_file, or if all three are None the full
-            source domain will be used.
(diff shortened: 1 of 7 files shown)


sup3r/pipeline/forward_pass.py
```diff
@@ -770,17 +727,28 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
...
                                             self.t_enhancements,
                                             self.spatial_pad,
                                             self.temporal_pad)
 
-        logger.info('Initializing ForwardPassStrategy for '
-                    f'{self.input_file_info}. Using n_nodes={self.nodes} with '
+        self.preflight()
+
+    @property
+    def worker_attrs(self):
```
grantbuster, line 741: This seems like a bad use of a property. Properties enable encapsulated compute. Whats the reason to have a hard-coded property value?

sup3r/preprocessing/data_handling.py
```diff
@@ -556,6 +556,10 @@ def __init__(self, file_paths, features, target=None, shape=None,
             files do not all have time indices or if there are few input files
             this should be set to one.
         """
+
+        msg = 'No files provided to DataHandler. Aborting.'
+        assert file_paths is not None and file_paths != [], msg
```
grantbuster, line 561: probably safer to do `assert file_paths is not None and bool(file_paths)`? I dont actually think you need the bool() call here but imo it's easier to read that way.

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: Yeah no reason. I guess better as a method?

grantbuster: No no my point was just make it an attribute! self._protected attribute if you don't want it messed with. Doesnt make sense to have any sort of dynamic code return a hard-coded value.

bnb32: Ah ok. I made it a property because its abstracted in InputMixIn and defined in both forward pass and data handling. I added an underscore :)

grantbuster: Lol why is it a property anywhere though??? Why wouldn't we just set `self._worker_attrs = [...]` in all of the init's? Hardcoded properties don't make sense - abstracted upstream or not.

grantbuster (approved):

PR #110, opened 2022-11-14
Bnb/async fwp

bnb32:
Enables running ForwardPass object initialization and ForwardPass.run_chunk() concurrently.


sup3r/pipeline/forward_pass.py
```diff
@@ -1705,24 +1728,36 @@ def incremental_check_run(cls, strategy, node_index, chunk_index):
...
+            Either a ForwardPass object, a concurrent future, or None. This
+            comes from either a direct ForwardPass initialization or a
+            ThreadPoolExecutor.submit() call.
+        """
+        if isinstance(fwp, ForwardPass):
+            fwp.run_chunk()
+        elif fwp is not None:
+            fwp = fwp.result()
+            if fwp is not None:
                 fwp.run_chunk()
```
grantbuster, line 1760: This is some weird double "if fwp is not None" logic... What do you think about this?
```
if isinstance(fwp, future):  # need to find future class from cf
    fwp = fwp.result()
if isinstance(fwp, ForwardPass):
    fwp.run_chunk()

File: sup3r/pipeline/forward_pass.py [hunk 2]
```diff
@@ -1738,61 +1773,81 @@ def run(cls, strategy, node_index):
...
+                for i, chunk_index in enumerate(
+                        strategy.node_chunks[node_index]):
+                    load_future = exe.submit(cls.incremental_check_load,
+                                             strategy=strategy,
+                                             node_index=node_index,
+                                             chunk_index=chunk_index)
+                    load_futures[load_future] = chunk_index
+                    fwp_future = exe.submit(cls._run_fwp_or_future,
+                                            load_future)
+                    fwp_futures[fwp_future] = chunk_index
```
grantbuster, line 1819: This is an interesting implementation, my only concern is how do you ensure that the 2 workers are working on a load future and a fwp future in parallel? This isn't enforced is it? I would think you could get scenarios where you get the 2 workers doing only 2x loads or only 2x fwps, thereby not really balancing the IO with the compute and possibly running into memory constraints.

grantbuster (changes requested):

sup3r/pipeline/forward_pass.py
bnb32: Yeah thats better

sup3r/pipeline/forward_pass.py
bnb32: Yeah my initial implementation ran exe.submit on fwp instantiations and returned those futures. Another thread went through those futures and when one was completed ran the corresponding fwp.run_chunk(). This was a more explicit separation but was not nearly the speed up we see here. I've been using this to run all the wtk/seasia runs and havent seen a mem error yet

grantbuster: Gotcha, so current implementation actually does an implicit dependency graph so that when pass_workers=2 it will only work on a single load+fwp at any given time because concurrent futures can only march through the compute queue in order. Looks good!

grantbuster (approved):

PR #114, opened 2022-11-16
moved meta collection out of a for loop - doesnt need to happen for e…

grantbuster:
…very dataset


bnb32 (approved):
Good call. Had this on the todo list but forgot about it.

PR #117, opened 2022-11-18
Bnb/dev

bnb32:
Added check for forward pass output chunk size. I used a padding that put the output chunks above 1.5 GB and didnt find out the output was constant until after collection! C'mon tensorflow.

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -1767,6 +1767,12 @@ class ForwardPass:
         fwp = None
         check = (not strategy.chunk_finished(chunk_index)
                  and not strategy.failed_chunks)
+
+        if strategy.failed_chunks:
+            msg = 'A forward pass has failed. Aborting all jobs.'
+            logger.error(msg)
+            raise MemoryError(msg)
+
         if check:
             fwp = cls(strategy, chunk_index=chunk_index, node_index=node_index)
         return fwp
@@ -1876,7 +1882,7 @@ class ForwardPass:
         """This routine runs a forward pass on single spatiotemporal chunk.
         """
         msg = (f'Running forward pass for chunk_index={self.chunk_index}, '
-               f'node_index={self.node_index}, file_paths={self.file_paths}.'
+               f'node_index={self.node_index}, file_paths={self.file_paths}. '
                f'Starting forward pass on chunk_shape={self.chunk_shape} with '
                f'spatial_pad={self.strategy.spatial_pad} and temporal_pad='
                f'{self.strategy.temporal_pad}.')
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -664,6 +664,7 @@ class ForwardPassStrategy(InputMixIn):
         self.pass_workers = pass_workers
         self.exo_kwargs = exo_kwargs or {}
         self.incremental = incremental
+        self._failed_chunks = False
         self._input_handler_class = None
         self._input_handler_name = input_handler
         self._max_nodes = max_nodes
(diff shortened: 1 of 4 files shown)


sup3r/pipeline/forward_pass.py
```diff
@@ -736,8 +737,26 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
                                             self.spatial_pad,
                                             self.temporal_pad)
 
+        msg = ('The average memory of a forward pass output chunk is '
+               f'{self.output_chunk_mem / 1e9:.2f} GB. Exceeding 1.5 GB could'
+               ' result in constant model output.')
+        if self.output_chunk_mem / 1e9 > 1.5:
+            logger.warning(msg)
+            warnings.warn(msg)
```
grantbuster, line 745: I have mixed feelings about this, do we know what's actually causing this? Is this a documented issue? I can't find anything online. 

Most importantly, the size of the output chunk at 1.5 GB is likely not the cause of this problem. It's almost certainly a memory overflow mid-network where the tensor is much larger. If we can confirm that tensorflow basically does a silent OOM error, we should just include a psutil check in the model.generate() method between layers.

grantbuster (changes requested):
I want to chat more about the OOM tensor issue...

Also, unrelated, can you change this log message to INFO? It's the most useful monitoring message IMO and should be always present in the logs: https://github.com/NREL/sup3r/blob/main/sup3r/pipeline/forward_pass.py#L1829

sup3r/pipeline/forward_pass.py
bnb32: I haven't been able to find anything either. I guess you're right that its probably not dependent on output size but some intermediate size. I'd prefer we do a check before rather than waiting to see if the output is bad. It's very clear that it's related to memory though since just increasing the padding passed a threshold gives constants all of a sudden.

grantbuster: But if you do the check before the first forward pass you're assuming A) this threshold you've set is going to be applicable to any model architecture (bad assumption) and B) the user is going to be parsing the log files for warning statements (unlikely if you're running over night). 

If you raise an exception in response to a bad chunk output you will 1) save a lot of compute time because the job won't continue to run, 2) save the human time of having to monitor a huge log file for really bad messages, and 3) have a check that works for any model architecture.

bnb32: Yeah I definitely agree on the first part. Im hoping there is a model agnostic pre check we could do. If we do a check on output then the only compute were avoiding is the h5 write.

grantbuster: I think there's a misunderstanding - we should check every chunk output, so the exception would be raise after the first forward pass chunk is complete and all other chunks would not be run.

bnb32: Yeah this makes sense. I was thinking that not every chunk was guaranteed to fail but nearly all will, except possibly the ones on the edges.

sup3r/pipeline/forward_pass.py
```diff
@@ -737,25 +738,18 @@ def __init__(self, file_paths, model_kwargs, fwp_chunk_shape,
...
-        n_elements *= (self.fwp_chunk_shape[1] + 2 * self.spatial_pad)
-        n_elements *= (self.fwp_chunk_shape[2] + 2 * self.temporal_pad)
-        n_elements *= len(self.output_features)
-        return 4 * n_elements
+    def failed_chunks(self):
+        """Check whether any forward passes have generated constant output."""
+        return self._failed_chunks
+
+    @failed_chunks.setter
+    def failed_chunks(self, failed):
```
grantbuster, line 749: This property+setter is functionally identical to just having a public self.failed_chunks attribute without the methods, any reason for doing the property+setter?

bnb32: Haha just figured we should make it clear there is this dedicated thing to check for failures rather than adding to the big list of attrs in init.

grantbuster: Gotcha, sounds good!

grantbuster (approved):

PR #118, opened 2022-11-18
Fix Sup3rGan interface

malihass:
Adding additional abstract class that contains training functions that can be used by other types of models such as conditional moment models.
Remove `Sup3rGen` which was only useful in the context of conditional moment learning. This class will be replaced by `Sup3rCondMom` when conditional moments models are ready to be uploaded.


grantbuster:
@malihass I like this a lot better, what do you think? Looks like it turned out to be mostly renaming/organizing but i think it's a more logical organization. 

One comment: Would you want to remove `meta`, `training_features`, and `output_features` from AbstractSingleModel? These are in AbstractInterface and I don't see a case where you'd use the SingleModel without the Interface. You could also do AbstractSingleModel as a subclass but i kind of like them separate, seems more modular.

malihass:
Ah good catch, I forgot to remove them. I had to add them because I had other old methods that I had included in the SingleModel one before and eventually decided against it.
I prefer to have the two abstract classes separate for now in case we need these methods separately.

grantbuster (approved):

PR #126, opened 2022-12-12
Conditional moment calculation

malihass:
- Creating the model `Sup3rCondMom` and `WindCondMom` which do not include the discriminator details
- Batch handler provides `output` and `mask` on top of`low_res` `high_res`. This is useful because sometime the `high_res` may not be directly obtainable from the output we are learning. The mask is useful if padding needs to be used.
- Different batch handler subclasses are implemented to learn conditional moments in different ways
    -  SF: learns the conditional moment of the sub filter velocity SF=(HR-LR)
    - Mom1: learns first conditional moment (HR|LR) or (SF|LR)
    - Mom2: learns second conditional moment ((HR - (HR|LR))^2 | LR) or ((SF - (SF|LR))^2 | LR) 
    - Mom2Sep: Learn second moment independently of the first moment (HR^2|LR) or (SF^2|LR)


sup3r/preprocessing/batch_handling.py
```diff
@@ -1379,7 +1838,10 @@ def __next__(self):
                 output_features_ind=self.output_features_ind,
                 training_features=self.training_features,
                 smoothing=self.smoothing,
-                smoothing_ignore=self.smoothing_ignore)
+                smoothing_ignore=self.smoothing_ignore,
+                model_mom1=self.model_mom1,
```
grantbuster, line 1842: @bnb32 at some point we need to start breaking these files up we don't need to do this right now. Typical guidance is <1k lines per file.

tests/test_train_wind_conditional_moments.py
```diff
@@ -0,0 +1,111 @@
...
+
+    x = np.random.uniform(0, 1, (4, 30, 30, 3))
+    hi_res_topo = np.random.uniform(0, 1, (60, 60))
+
+    y = model.generate(x, exogenous_data=(None, hi_res_topo))
+
+    assert y.shape[0] == x.shape[0]
+    assert y.shape[1] == x.shape[1] * 2
+    assert y.shape[2] == x.shape[2] * 2
+    assert y.shape[3] == x.shape[3] - 1
```
grantbuster, line 111: @malihass thanks for writing tests!

grantbuster (changes requested):
Looks good overall, only big change request is to fix all of the copy-paste code.

sup3r/models/wind_conditional_moments.py
```diff
@@ -0,0 +1,335 @@
...
+        output_features = kwargs['output_features']
+        msg = ('Last output feature from the data handler must be topography '
+               'to train the WindCC model, but received output features: {}'
+               .format(output_features))
+        assert output_features[-1] == 'topography', msg
+        output_features.remove('topography')
+        kwargs['output_features'] = output_features
+        super().set_model_params(**kwargs)
+
+    def _reshape_norm_topo(self, hi_res, hi_res_topo, norm_in=True):
```
bnb32, line 46: Similar comments here as on conditional_moments.py - We should move lots of this code into a super class since WindGan can inherit much of it as well. Maybe call the super class BaseWindInterface or something.

sup3r/utilities/plot_utilities.py
```diff
@@ -0,0 +1,384 @@
...
+
+import imageio
+import matplotlib
+from matplotlib import cm
+import matplotlib.pyplot as plt
+import numpy as np
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+
+
+def prettyLabels(xlabel, ylabel, fontsize=14, title=None):
```
bnb32, line 15: should use underscore for function names with multiple words, as per python naming conventions

sup3r/preprocessing/batch_handling.py
```diff
@@ -90,6 +112,71 @@ def reduce_features(high_res, output_features_ind=None):
         else:
             return high_res[..., output_features_ind]
 
+    # pylint: disable=W0613
+    @staticmethod
```
bnb32, line 116: Im confused by this method. Why do we need this for just a high_res pass-through?

sup3r/preprocessing/batch_handling.py
```diff
@@ -70,6 +82,16 @@ def high_res(self):
         """Get the high-resolution data for the batch."""
         return self._high_res
 
+    @property
```
bnb32, line 85: Can you explain in the doc string how output is different from high_res?

sup3r/preprocessing/batch_handling.py
```diff
@@ -1112,14 +1453,123 @@ def __next__(self):
...
+                s_padding=self.s_padding,
+                t_padding=self.t_padding)
 
             self._i += 1
             return batch
         else:
             raise StopIteration
 
 
+class ValidationDataMom1SF(ValidationData):
```
bnb32, line 1467: Maybe we could move some of this to a new file for conditional_moment_batch_handling

sup3r/preprocessing/batch_handling.py
```diff
@@ -32,10 +34,10 @@
 
 
 class Batch:
```
bnb32, line 36: I think we should be subclassing for most of the changes in here - instead of editing the older classes. You could implement these new classes by subclassing in another file called conditional_moment_batch_handling.py, or something.

bnb32:
> @bnb32 at some point we need to start breaking these files up we don't need to do this right now. Typical guidance is <1k lines per file.  

Yeah agreed. That is one of my comments on this PR - also just got to that part of "clean code" :)

bnb32 (changes requested):
I have many similar comments to Grant apparently, and maybe some different ones. Looks good overall just needs some changes with structure.

sup3r/preprocessing/batch_handling.py
bnb32: @grantbuster Do you agree with this^

sup3r/utilities/plot_utilities.py
grantbuster: snake_case

sup3r/preprocessing/batch_handling.py
grantbuster: I could go either way, pro: I could see the value in adding the masking logic to the base class, con: adds bloat and overhead compute that may not be used.

sup3r/preprocessing/batch_handling.py
```diff
@@ -90,6 +112,71 @@ def reduce_features(high_res, output_features_ind=None):
...
+        output_features_ind : list | np.ndarray | None
+            List/array of feature channel indices that are used for generative
+            output, without any feature indices used only for training.
+        """
+        return high_res
+
+    # pylint: disable=E1130
+    @staticmethod
+    def make_mask(high_res, s_padding=None, t_padding=None):
+        """Make mask for output
```
bnb32, line 147: Can you say more about what the mask is used for and why? The doc string here doesn't say much more than the function name ;)

sup3r/preprocessing/batch_handling.py
bnb32: Maybe the masking could go into the old classes and the cond moment/output stuff could be subclassed.

sup3r/preprocessing/batch_handling.py
```diff
@@ -155,11 +253,203 @@ def get_coarse_batch(cls, high_res,
...
+        return (high_res - enhanced_lr - out)**2
+
+
+class BatchMom2SepSF(Batch):
+    """Batch of low_res, high_res and output data
+    when learning second moment of subfilter vel
+    separate from first moment"""
+
+    @staticmethod
+    def make_output(low_res, high_res,
```
bnb32, line 380: This is just BatchMom1SF.make_output()**2 right? Maybe this should just use that method.

sup3r/preprocessing/batch_handling.py
```diff
@@ -155,11 +253,203 @@ def get_coarse_batch(cls, high_res,
...
+        output_features_ind : list | np.ndarray | None
+            List/array of feature channel indices that are used for generative
+            output, without any feature indices used only for training.
+        """
+        # Remove first moment from HR and square it
+        out = model_mom1._tf_generate(low_res).numpy()
+        return (high_res - out)**2
+
+
+class BatchMom2Sep(Batch):
```
bnb32, line 301: This could also just call Batch.make_output()**2

malihass: This one might make the code less readable, no?

bnb32: Well I think Batch should be subclassed as BatchMom1Sep (or something like that) with make_output returning just high_res and then BatchMom2Sep could just return BatchMom1Sep.make_output**2, which I think reads better than having the same repeated doc string.

bnb32: Even better would be a single n-th conditional moment Batch class that is completely general, which just takes n as an input. But maybe this isn't feasible.

sup3r/models/conditional_moments.py
```diff
@@ -0,0 +1,850 @@
...
+            msg = ('Something went wrong. Tried to load a custom network '
+                   'but ended up with a model of type "{}"'
+                   .format(type(model)))
+            logger.error(msg)
+            raise TypeError(msg)
+
+        return model
+
+    @property
+    def means(self):
```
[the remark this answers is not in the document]
malihass: I reorganized `abstract`, `conditionalMoment` and `base`. Most duplicate code is gone now.  I did leave `meta` since it is an abstract method which needs to be redefined. `model_params` change a little bit between Sup3rCondMom and Sup3rGan, so I left it as well. All the normalization stuff is in the abstractSingleModel now. I think it also makes sense to do this since any future model will presumably need these normalization methods. I'll leave this unresolved for now until you validate.

sup3r/models/conditional_moments.py
```diff
@@ -0,0 +1,850 @@
...
+        if isinstance(self.history, pd.DataFrame):
+            fp_history = os.path.join(out_dir, 'history.csv')
+            self.history.to_csv(fp_history)
+
+        self.save_params(out_dir)
+
+        logger.info('Saved model to disk in directory: {}'.format(out_dir))
+
+    @classmethod
+    def load(cls, model_dir, verbose=True):
```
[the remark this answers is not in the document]
malihass: I agree, there are too many duplicates. I put these redundancies between `Sup3rCondMom` and `Sup3rGan` in the `AbstractSingleModel` class. Would that work?

sup3r/preprocessing/batch_handling.py
malihass: I think that for now I will add the masking only for the conditional moment batching system. We can decide later if we want to add this for all batching :-)

sup3r/models/conditional_moments.py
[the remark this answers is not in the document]
grantbuster: Nice looks great and pylint says no duplicate code now. Thanks!

grantbuster (approved):
@malihass, everything looks good on my end. Shame about the `_tf_generate()` vs. `_tf_generate_wind()` methods, i'm assuming you did this because of the multi-class inheritance conflict? Good solution but still a shame it couldnt be cleaner haha. 

I'm approving but I still want @bnb32 to get a chance to look at the updates to the batch handling before you merge.

sup3r/preprocessing/conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,854 @@
...
+            Spatial enhancement factor
+        t_enhance : int | None
+            Temporal enhancement factor
+        model_mom1 : Sup3rCondMom | None
+            Model used to modify the make the batch output
+        output_features_ind : list | np.ndarray | None
+            List/array of feature channel indices that are used for generative
+            output, without any feature indices used only for training.
+        """
+        return BatchMom1.make_output(low_res, high_res,
```
bnb32, line 340: I would make this super().make_output instead of BatchMom1.make_output()

sup3r/preprocessing/conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,854 @@
...
+        out = model_mom1._tf_generate(low_res).numpy()
+        enhanced_lr = spatial_simple_enhancing(low_res,
+                                               s_enhance=s_enhance)
+        enhanced_lr = temporal_simple_enhancing(enhanced_lr,
+                                                t_enhance=t_enhance)
+        enhanced_lr = Batch.reduce_features(enhanced_lr, output_features_ind)
+        return (high_res - enhanced_lr - out)**2
+
+
+class BatchMom2SepSF(BatchMom1):
```
bnb32, line 385: I would subclass BatchMom1SF and then use super().make_output

bnb32:
> @malihass, everything looks good on my end. Shame about the `_tf_generate()` vs. `_tf_generate_wind()` methods, i'm assuming you did this because of the multi-class inheritance conflict? Good solution but still a shame it couldnt be cleaner haha.
> 
> I'm approving but I still want @bnb32 to get a chance to look at the updates to the batch handling before you merge.

Is there really not a way to override that method with multi class inheritance?

sup3r/models/abstract.py
```diff
@@ -424,3 +657,290 @@ def early_stop(history, column, threshold=0.005, n_epoch=5):
...
+
+        if extras is not None:
+            for k, v in extras.items():
+                self._history.at[epoch, k] = v
+
+        return stop
+
+
+# pylint: disable=E1101,W0201,E0203
+class AbstractWindInterface(ABC):
```
bnb32, line 750: Can't we have this subclass AbstractInterface and then override the methods instead of appending the _wind suffix? @grantbuster

sup3r/models/wind_conditional_moments.py
```diff
@@ -0,0 +1,215 @@
...
+import tensorflow as tf
+
+from sup3r.models.abstract import AbstractWindInterface
+from sup3r.models.conditional_moments import Sup3rCondMom
+
+
+logger = logging.getLogger(__name__)
+
+
+class WindCondMom(Sup3rCondMom, AbstractWindInterface):
```
bnb32, line 14: What's the reason for defining tf_generate_wind and then using it to define tf_generate? Same for generate_wind, and set_model_params_wind? I think we could just override the standard names in this class

bnb32 (changes requested):
This looks great. Thanks! I just have some questions about the inheritance stuff and whether we can avoid the _wind suffix work around.

sup3r/models/abstract.py
grantbuster: Related to `_tf_generate()` vs. `_tf_generate_wind()`? You could move `_tf_generate()` to the `AbstractSingleModel` and have `AbstractWindInterface(AbstractSingleModel)` but then you're trading an inheritance conflict for the diamond problem... I think i'd prefer the inheritance conflict handled explicitly rather than the diamond problem.

bnb32: The methods could also just be moved into WindCondMom instead of AbstractWindInterface

grantbuster: I don't think so, the point of the `AbstractWindInterface` is to have a common mixin class for the wind models - `WindGan` and `WindCondMom`. If you put the methods in `WindCondMom` you wouldn't have them in `WindGan`

bnb32: Oh yeah I missed them in WindGan. Ok this is fine then!

bnb32 (approved):
_wind seems like the best solution since WindGan also inherits this stuff. Good work!

sup3r/models/abstract.py
grantbuster: Yeah it's a messy web haha. I think this is the best of the not-great options.

sup3r/models/wind_conditional_moments.py
malihass: So `Sup3rCondMom` already has a `tf_generate` method that we need to redefine to handle the topography. We have 2 choices, either I redefine `tf.generate` in `WindCondMom` and `WindGan` but this create redundancy since `tf_generate` is the same for both. The alternative is what I did but then I have to refer to this method with a different name, otherwise WindCondMom and WindGan would not know which one to grab.

bnb32: Yeah I missed it in WindGan. This is best option. Just those little things I mentioned in the batch handling and you're good to go!

PR #127, opened 2022-12-20
Gb/bc cs ratio

grantbuster:


sup3r/bias/bias_calc.py
```diff
@@ -414,6 +390,109 @@ def get_base_data(base_fps, base_dset, base_gid, base_handler,
...
+        daily_reduction : str
+            Option to do a reduction of the hourly+ source base data to daily
+            data. Can be None (no reduction, keep source time frequency), "avg"
+            (daily average), "max" (daily max), or "min" (daily min)
+
+        Returns
+        -------
+        base_data : np.ndarray
+            1D array of base data spatially averaged across the base_gid input
+            and possibly daily-averaged as well.
```
bnb32, line 472: I suppose this could be daily-averaged or max/min. Everything looks good so I had to find one thing to nitpick.

bnb32 (approved):

sup3r/bias/bias_calc.py
grantbuster: I think you're saying it should say min/max'd in addition to daily-averaged? Makes sense to me! added along with a real functional fix.

PR #128, opened 2022-12-21
Bnb/dev

bnb32:
Some refactoring to reduce length of class attribute lists in __init_ for DataHandler and ForwardPassStrategy. Replaced some for loops with vectorized equivalents. This gave us a few seconds for each fwp class.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -61,15 +61,10 @@ disable=
         unpacking-in-except,
         old-raise-syntax,
         backtick,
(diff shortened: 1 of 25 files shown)


grantbuster (approved):
This all looks good but I definitely want to encourage us to stop making changes to all the argument names constantly haha. This is still a good change in that spirit because it's more future proof than having a bunch of separate args floating around though.

PR #130, opened 2022-12-22
Bnb/visual qa

bnb32:
Centralizing cli methods. Added simple visual-qa module.


sup3r/cli/base.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseCLI:
+    """Base CLI class used to create CLI for modules in ModuleName"""
+
+    @classmethod
+    def from_config(cls, module_name, module_class, ctx, config_file, verbose):
```
grantbuster, line 28: What all is preventing this from being used everywhere? The multi-node jobs? Could the first part through line ~83 be put into a "get_from_config_parameters()" method that could be used everywhere? 

I'm assuming this first part is basically copied from cli to cli but the multi-node submission is unique in fwp, solar, and bias.

sup3r/qa/feature_plots.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, file_paths, out_pattern, features, time_step=10,
+                 workers=None, **kwargs):
+        """
+        Parameters
+        ----------
+        file_paths : list | str
+            Specifies the files to use for the plotting routine. This is either
+            a list of h5 files generated by the forward pass module or a string
+            pointing to h5 forward pass output which can be parsed by glob.glob
+        out_pattern : str
```
grantbuster, line 32: Should this include a "*.png"? Might be useful to include in the docstring.

sup3r/qa/feature_plots.py
```diff
@@ -0,0 +1,211 @@
...
+        self.workers = workers
+        self.kwargs = kwargs
+
+    def run(self):
+        """
+        Create plot figures for all the features in self.features. For each
+        feature there will be n_files created, where n_files is the number of
+        timesteps in the h5 files provided divided by self.time_step.
+        """
+        with MultiFileResourceX(self.file_paths) as res:
```
grantbuster, line 64: I would recommend that the ResourceClassHandler be a kwarg. We almost never use the "*X" resource handlers actually. The "extraction" handlers mostly include extra features for external users who want to query based on lat/lon and whatnot. They are not subclasses of the `MultiFileResource` handler (for example) and do not have as many useful features for our use cases.

sup3r/qa/feature_plots.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        start_time = time_index[t_slice.start]
+        stop_time = time_index[t_slice.stop - 1]
+        logger.info(f'Plotting time average for {feature} from '
+                    f'{start_time} to {stop_time}.')
+        fig = plt.figure()
+        title = f'{feature}: {start_time} - {stop_time}'
+        plt.suptitle(title)
+        plt.scatter(res.meta.longitude, res.meta.latitude,
+                    c=np.mean(res[feature][t_slice], axis=0), **self.kwargs)
```
grantbuster, line 155: Recommendations:

1. Add option for a spatial slice that would plot at an interval. For 2km over CONUS you can often get away with plotting every 5 or 10 pixels which speeds things up significantly. 
2. The extraction from the rex resource handlers should look like `res[feature, t_slice, s_slice]` where what you have `res[feature][t_slice]` will actually retrieve the FULL feature array and then subsequently slice it in memory. This has been a big debate because xarray and h5py use the syntax you're using here but it's too big of a refactor right now given how widely rex is being used across the lab.

grantbuster (changes requested):

sup3r/cli/base.py
bnb32: Yeah data-extract and data-collect are unique as well but good call on extracting the first bit.

sup3r/qa/feature_plots.py
bnb32: Ah gotcha, good call.

sup3r/qa/feature_plots.py
bnb32: Ah yeah, keep using the xarray syntax by default. Good catch.

sup3r/bias/bias_calc_cli.py
```diff
@@ -42,35 +36,15 @@ def main(ctx, verbose):
...
-    ctx.obj['OUT_DIR'] = status_dir
-    config = BaseConfig(config_file)
-    config['status_dir'] = status_dir
-    config_verbose = config.get('log_level', 'INFO')
-    config_verbose = (config_verbose == 'DEBUG')
-    verbose = any([verbose, config_verbose, ctx.obj['VERBOSE']])
-
-    init_mult('sup3r_bias_calc', os.path.join(status_dir, 'logs/'),
-              modules=[__name__, 'sup3r'], verbose=verbose)
+    config = BaseCLI.from_config_preflight(ModuleName.BIAS_CALC, ctx,
```
grantbuster, line 39: Nice this looks great!

grantbuster (approved):

PR #132, opened 2022-12-23
Parallel optimization

grantbuster:


bnb32 (approved):
Looks good! Glad we finally figured this out and didn't leave all that free speedup on the table.

PR #138, opened 2023-01-09
Gb/multi gpu

grantbuster:


grantbuster:
@bnb32 can you take a look at what might be going on with the data-centric test? I have no idea how my edits would affect that code 

https://github.com/NREL/sup3r/actions/runs/3876667898/jobs/6610776941

sup3r/models/abstract.py
```diff
@@ -761,22 +778,157 @@ def finish_epoch(self, epoch, epochs, t0, loss_details,
...
+
+            with ThreadPoolExecutor(max_workers=len(self.gpu_list)) as exe:
+                for i in range(len(self.gpu_list)):
+                    futures.append(exe.submit(self.get_single_grad,
+                                              lr_chunks[i],
+                                              hr_true_chunks[i],
+                                              training_weights,
+                                              device_name=f'/gpu:{i}',
+                                              **calc_loss_kwargs))
+            for i, future in enumerate(futures):
```
bnb32, line 895: why don't we need as_completed() here?

sup3r/models/abstract.py
```diff
@@ -960,3 +1113,54 @@ def _tf_generate_wind(self, low_res, hi_res_topo):
                 raise RuntimeError(msg) from e
 
         return hi_res
+
+    @tf.function()
```
bnb32, line 1117: having this copied code hurts me. but alas...

sup3r/models/abstract.py
```diff
@@ -761,22 +778,157 @@ def finish_epoch(self, epoch, epochs, t0, loss_details,
...
+                hi_res_gen = self._tf_generate(low_res)
+                loss_out = self.calc_loss(hi_res_true, hi_res_gen,
+                                          **calc_loss_kwargs)
+                loss, loss_details = loss_out
+
+                grad = tape.gradient(loss, training_weights)
+
+        return grad, loss_details
+
+    def run_gradient_descent(self, low_res, hi_res_true, training_weights,
```
bnb32, line 830: this is a nice simple solution. well done!

sup3r/models/abstract.py
bnb32: Also, applying the different gradients like this is better than applying the average of the gradients once?

bnb32 (approved):
Wish we could restructure to avoid duplicating some code but overall an elegant solution to the multi-gpu problem.

bnb32:
Wish we could restructure to avoid duplicating some code but overall an elegant solution to the multi-gpu problem.

sup3r/models/abstract.py
grantbuster: we could use as_completed() but the futures will finish at basically the same time so i was just keeping it simple. And i dont think either is better - averaging with a single weight update might actually be more intuitive, but this is nice and simple.

PR #141, opened 2023-01-11
Bnb/dev

bnb32:
some refactoring - ForwardPassStrategy/Regridder -> DistributedProcess. Changed inheritance order in wind models to remove _wind methods.


sup3r/models/abstract.py
```diff
@@ -909,9 +909,8 @@ class AbstractWindInterface(ABC):
     Abstract class to define the required training interface
     for Sup3r wind model subclasses
     """
-
-    @staticmethod
-    def set_model_params_wind(**kwargs):
+    # pylint: disable=E0211
+    def set_model_params(self, **kwargs):
```
grantbuster, line 913: shouldnt this be a static method still?

sup3r/models/wind.py
```diff
@@ -12,7 +12,7 @@
 logger = logging.getLogger(__name__)
 
 
-class WindGan(Sup3rGan, AbstractWindInterface):
+class WindGan(AbstractWindInterface, Sup3rGan):
```
grantbuster, line 15: oh nice! just reordering solves the inheritance priority? Great!!

sup3r/utilities/regridder.py
```diff
@@ -499,6 +496,10 @@ def __init__(self, source_files, out_pattern, target_meta, heights,
                                        cache_pattern=cache_pattern,
                                        n_chunks=n_chunks,
                                        max_workers=self.query_workers)
+        DistributedProcess.__init__(self, max_nodes=max_nodes,
```
grantbuster, line 499: Oh wow i didnt know you could just call the inherited init directly and pass in self! that makes so much sense. Nice.

sup3r/utilities/utilities.py
```diff
@@ -26,6 +26,130 @@
 logger = logging.getLogger(__name__)
 
 
+class DistributedProcess:
```
grantbuster, line 29: Suggest this goes into `utilities/execution.py` or similar.... the utilities.py file is getting totally ridiculous.

grantbuster (changes requested):

sup3r/models/wind.py
bnb32: Yeah! Python's solution to the diamond problem!

sup3r/utilities/utilities.py
bnb32: Yeah I figured you'd have a suggestion for where to put this. I didn't like it here either.

sup3r/models/abstract.py
bnb32: Yep lol

sup3r/utilities/regridder.py
bnb32: Thanks! Didn't know this worked either just figured it _should_ lol

grantbuster (approved):

PR #142, opened 2023-01-19
Gb/trh loss

grantbuster:


sup3r/preprocessing/data_handling.py
```diff
@@ -1106,7 +1106,7 @@ def preflight(self):
         msg = (f'sample_shape[2] ({self.sample_shape[2]}) cannot be larger '
                'than the number of time steps in the raw data '
                f'({len(self.raw_time_index)}).')
-        if len(self.raw_time_index) >= self.sample_shape[2]:
+        if len(self.raw_time_index) < self.sample_shape[2]:
```
bnb32, line 1109: good catch, fixed this in my current dev branch also.

bnb32, line 1109: good catch, fixed this in my current dev branch also.

sup3r/models/surface.py
```diff
@@ -104,34 +119,34 @@ def __len__(self):
...
-        model = cls(features, s_enhance, **kwargs)
+        fp_params = os.path.join(model_dir, 'model_params.json')
+        assert os.path.exists(fp_params), f'Could not find: {fp_params}'
+        with open(fp_params, 'r') as f:
+            params = json.load(f)
+
+        meta = params['meta']
+        args = signature(cls.__init__).parameters
+        kwargs = {k: v for k, v in meta.items() if k in args}
+        model = cls(**kwargs)
```
bnb32, line 149: you use this code block in linear.py also, and it seems useful as a separate function or an inherited method?

bnb32 (changes requested):

sup3r/preprocessing/data_handling.py
grantbuster: it was driving me insane haha

sup3r/models/surface.py
grantbuster: Yeah good point i think we can actually inherit the surface model from the linear model. Let me test it.

bnb32 (approved):

PR #145, opened 2023-03-27
Gb/exo cache

grantbuster:


bnb32 (changes requested):
Worth adding a test for the exo caching?

grantbuster:
> Worth adding a test for the exo caching?

Yep, added!

PR #146, opened 2023-04-06
Bnb/era edits

bnb32:
DataHandlerNCforERA addition. resolved issues with pandas version update. unpinned xarray, netcdf, sphinx. Incremented version. Removed support for python 3.7. Added python 3.10.


sup3r/postprocessing/file_handling.py
```diff
@@ -119,6 +119,29 @@ class OutputMixIn:
...
     """
 
+    @staticmethod
+    def get_time_dim_name(filepath):
+        """Get the name of the time dimension in the given file
+
+        Parameters
+        ----------
+        filepath : str
+            Path to the file
```
grantbuster, line 129: .nc file only? Move this into the NC specific class?

sup3r/preprocessing/data_handling.py
```diff
@@ -408,6 +407,44 @@ def target(self, target):
...
+        time_key : str
+            Name of the time dimension in the given file
+        """
+        with xr.open_dataset(filepath) as handle:
+            valid_vars = set(handle.dims)
+            time_key = list({'time', 'Time'}.intersection(valid_vars))
+        if len(time_key) > 0:
+            return time_key[0]
+        else:
+            return 'time'
```
grantbuster, line 446: why is this a class method instead of static? Can you make a utility function for this since the same functionality is needed in post processing above? if not, move to the nc specific classes? Can you update the docstring to clarify that this is just for .nc?

sup3r/utilities/utilities.py
```diff
@@ -1011,8 +1011,11 @@ def calc_height(data, raster_index, time_slice=slice(None)):
         del gp
 
     elif all(field in data for field in ('zg', 'orog')):
+        if len(data['orog'].dims) == 3:
+            hgt = data['orog'][(0,) + tuple(raster_index)]
+        else:
+            hgt = data['orog'][tuple(raster_index)]
```
grantbuster, line 1017: These interp utilities are nearing the point where we should move them all into an interp utilities class. Complexity for this method alone is quite high, and the utilities.py file alone is massive

sup3r/utilities/utilities.py
```diff
@@ -1093,6 +1096,7 @@ def interp_to_level(var_array, lev_array, levels):
         raise RuntimeError(msg)
 
     nans = np.isnan(lev_array)
+    logger.info('lev_array.shape: {}'.format(lev_array.shape))
```
grantbuster, line 1099: seems debug worthy

grantbuster (changes requested):
looks good, minor comments

sup3r/utilities/utilities.py
bnb32: Yeah agreed. Moved to new class.

grantbuster (approved):

PR #149, opened 2023-05-12
Bug fix for conditional moment with topography

malihass:
The computation of the second moment when using exogenous data was not working when we need to call the first moment in the batch handler. The batch handler needs to call the first moment with low res and the exogenous data rather than just the low res. 

- Created separate batch handler for conditional moment with exogenous data
- Added test suite to test for the new classes added


README.rst
```diff
@@ -64,6 +64,7 @@ Option 2: Clone repo (recommended for developers)
     6) Install ``sup3r`` and its dependencies by running:
        ``pip install .`` (or ``pip install -e .`` if running a dev branch
        or working on the source code)
+    7) Set up the pre-commit hooks by calling ``pre-commit install``
```
bnb32, line 67: We should use `pip install pre-commit` to be consistent.

sup3r/preprocessing/wind_conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,354 @@
...
+        t_enhance : int | None
+            Temporal enhancement factor
+        model_mom1 : Sup3rCondMom | None
+            Model used to modify the make the batch output
+        output_features_ind : list | np.ndarray | None
+            List/array of feature channel indices that are used for generative
+            output, without any feature indices used only for training.
+        t_enhance_mode : str
+            Enhancing mode for temporal subfilter.
+            Can be either constant or linear
```
bnb32, line 58: include Return description.

sup3r/preprocessing/wind_conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,354 @@
...
+        """
```
bnb32, line 105: include Return description

sup3r/preprocessing/wind_conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,354 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class WindBatchMom1(BatchMom1):
+    """Batch of low_res, high_res and output wind data"""
+
+
+class WindBatchMom1SF(WindBatchMom1):
+    """Batch of low_res, high_res and output wind data
```
bnb32, line 27: my OCD forces me to ask if we can only add a line break at the point the text would go passed the 79th column.

sup3r/preprocessing/wind_conditional_moment_batch_handling.py
```diff
@@ -0,0 +1,354 @@
...
+        t_enhance : int | None
+            Temporal enhancement factor
+        model_mom1 : Sup3rCondMom | None
+            Model used to modify the make the batch output
+        output_features_ind : list | np.ndarray | None
+            List/array of feature channel indices that are used for generative
+            output, without any feature indices used only for training.
+        t_enhance_mode : str
+            Enhancing mode for temporal subfilter.
+            Can be either 'constant' or 'linear'
```
bnb32, line 149: Return description

bnb32 (changes requested):

README.rst
malihass: Just to clarify, do we want to add pre-commit to the requirements.txt ?

bnb32: No I think we can leave this as an optional install.

grantbuster: maybe note that ``pre-commit install`` is optional too

malihass:
So after testing further, this PR only slows down the way the batches are assembled for the conditional moment model that use topography, which should be expected. @bnb32  I think we can consider this as ready to be re-reviewed  and we can revisit it at later times if needed. The execution time is the same when not using sub filter + topography

PR #151, opened 2023-06-01
Bnb/bc smoothing

bnb32:


sup3r/bias/bias_calc.py
```diff
@@ -581,12 +596,30 @@ def fill_extend(self, out, smooth_extend):
...
             for idt in range(self.NT):
-                arr[..., idt] = nn_fill_array(arr[..., idt])
+
+                arr_smooth = arr[..., idt]
+
+                needs_fill = (fill_extend or smooth_extend > 0
+                              or smooth_interior > 0)
+
+                if needs_fill:
+                    arr_smooth = nn_fill_array(arr_smooth)
```
grantbuster, line 606: Probably more accurate to do a `if np.isnan(arr_smooth).any() and fill_extend:` instead of the funky OR boolean

sup3r/bias/bias_transforms.py
```diff
@@ -64,6 +65,12 @@ def local_linear_bc(input, feature_name, bias_fp, lr_padded_slice,
         source shape will be used.
     out_range : None | tuple
         Option to set floor/ceiling values on the output data.
+    smoothing : float
```
grantbuster, line 68: why did you add this to the transforms module? Just so we don't have to recalc the bc factors? Or so you can test a bunch of smoothing factors at runtime? 

The docstring isn't accurate: no inside/outside spatial domain or threshold input here

sup3r/bias/bias_calc.py
```diff
@@ -553,9 +553,15 @@ def _run_single(cls, bias_data, base_fps, bias_feature, base_dset,
                                         bias_feature, base_dset)
         return out
 
-    def fill_extend(self, out, smooth_extend):
+    def fill_smooth_extend(self, out, fill_extend=True, smooth_extend=0,
```
grantbuster, line 556: nitpicky but fill and extend should be together in this title (we're filling the extension) or just "fill_and_smooth"

grantbuster (changes requested):

sup3r/bias/bias_transforms.py
bnb32: Yeah so we don't have to recalc, which also allows me to test. Wasn't sure if we should just keep the smoothing here and remove the other.

sup3r/bias/bias_calc.py
bnb32: Yeah I like either better lol

sup3r/bias/bias_transforms.py
grantbuster: One argument for only keeping the smoothing at the bias calc step: we'll be less likely to double-smooth. Consider that we will probably smooth the extrapolated extent and will be hard to not duplicate the smoothing in the transforms module. 

For testing i would advocate for just writing a script outside of the repo that duplicates the bias correction files with different smoothing factors.

bnb32: Agree on the double smoothing, although I'm not sure how much this matters outside the valid domain. I would argue for keeping the smoothing in the transforms since I don't see a reason to have to rerun bias-calc just to change smoothing. Keeping the stored bias correction factors as "true" as possible seems reasonable.

grantbuster: Yeah you just loose the ability to smooth inside the valid domain vs. outside... That's fine though if that's your preference... I'd say keep both because the ability to smooth inside vs. outside is important esp. for sup3rcc

sup3r/bias/bias_calc.py
```diff
@@ -581,12 +598,30 @@ def fill_extend(self, out, smooth_extend):
         for key, arr in out.items():
             nan_mask = np.isnan(arr[..., 0])
             for idt in range(self.NT):
-                arr[..., idt] = nn_fill_array(arr[..., idt])
+
+                arr_smooth = arr[..., idt]
+
+                needs_fill = (np.isnan(arr_smooth).any() and fill_extend
+                              or smooth_interior > 0)
```
grantbuster, line 605: You should be cautious with these kinds of operations, logical operators have an order of operations priority that is not intuitive. It's fine as-is but I would use parenthesis to make the order of operations explicit.

grantbuster (approved):

sup3r/bias/bias_calc.py
bnb32: Yeah good call, it's bitten me in the past.

PR #152, opened 2023-07-05
Bnb/dev

bnb32:
small edit to handle multi-step spatial only models

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -429,6 +429,8 @@ class ForwardPassSlicer:
             Low resolution slices to be enhanced
         enhancement : int
             Enhancement factor
+        step : int | None
+            Step size for slices
 
         Returns
         -------
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -1589,16 +1589,20 @@ class ForwardPass:
         i_lr_s : int
             Axis index for the low-resolution spatial_1 dimension
         """
-
+        current_model = None
         if exo_data is not None:
             for i, arr in enumerate(exo_data):
                 if arr is not None:
-                    current_model = (model if not hasattr(model, 'models')
-                                     else model.models[i])
-                    if current_model.input_dims == 4:
-                        exo_data[i] = np.transpose(arr, axes=(2, 0, 1, 3))
-                    else:
-                        exo_data[i] = np.expand_dims(arr, axis=0)
+                    if not hasattr(model, 'models'):
+                        current_model = model
+                    elif i < len(model.models):
+                        current_model = model.models[i]
+
+                    if current_model is not None:
+                        if current_model.input_dims == 4:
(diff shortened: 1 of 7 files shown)


sup3r/preprocessing/data_handling.py
```diff
@@ -264,11 +264,15 @@ def file_paths(self, file_paths):
...
         if isinstance(self._file_paths, str):
-            if '*' in self._file_paths:
+            if '*' in file_paths:
                 self._file_paths = glob.glob(self._file_paths)
             else:
                 self._file_paths = [self._file_paths]
 
+        msg = ('No valid files provided to DataHandler. '
+               f'Received file_paths={file_paths}. Aborting.')
+        assert len(self._file_paths) > 0 and file_paths is not None, msg
```
grantbuster, line 274: Probably better to check the None condition first. If file_paths is None, you'll get a weird error on cannot run `len(None)` instead of `is not None` i think?

grantbuster (approved):
looks good! one suggested edit.

sup3r/preprocessing/data_handling.py
bnb32: Yeah good call

sup3r/pipeline/forward_pass.py
```diff
@@ -1587,16 +1589,20 @@ def _reshape_data_chunk(model, data_chunk, exo_data):
         i_lr_s : int
             Axis index for the low-resolution spatial_1 dimension
         """
-
+        current_model = None
```
bnb32, line 1592: @grantbuster, what do you think about this addition? I added this so we could run with multi step spatial only models but I feel iffy about allowing len(exo_data) > len(model.models) in general.

grantbuster: hmmm you might need to explain to me how this enables multi step spatial only. How many exo arrays for how many models? Honestly we should refactor this at some point anyway and make it more explicit what exo data is going to what model. too much implicit logic that we have to manage right now.

bnb32: With two spatial models there was an indexing error - its needs 3 exo arrays but only has two models. When this is a spatial model and a temporal model it just needs 2 exo arrays.

grantbuster: Yeah i think this just points to the larger issue of implicit logic to handle these exo array mappings to models... Really we should make a dictionary of exo data in the config, e.g. {'exo_100km': {...}, 'exo_20km': {...}} and then have a mapping to models e.g. {'model0': [exo_100km, exo_20km], 'model3': [exo...]}.

If you have enough time to do this serious refactor i'd say go for it. If not, whatever logic fixes the problem is probably fine for now although temporary until we have the funds to do a better solution. 

Does that make sense? Thoughts?

bnb32: Yeah definitely makes sense. I can add the refactor to my list but lets implement a quicker fix for now.

grantbuster: Agreed!!

PR #161, opened 2023-07-30
Bnb/deterministic batching

bnb32:
Batch handler which uses deterministic sampling of the data handler data and precomputes + caches observations before training. 

TODO: should refactor base batch handler so that previously cached observations can be loaded from cache without loading data from data handlers. The caching of observations might also need to be done in batches or one at a time instead of all at once, due to mem limits.

### DIFF
--- a/sup3r/preprocessing/batch_handling.py
+++ b/sup3r/preprocessing/batch_handling.py
@@ -244,7 +244,8 @@ class ValidationData:
                                                         self.sample_shape[:2])
                     temporal_slice = uniform_time_sampler(h.val_data,
                                                           self.sample_shape[2])
-                    tuple_index = tuple([*spatial_slice, temporal_slice, np.arange(h.val_data.shape[-1])])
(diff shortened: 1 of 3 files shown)


malihass:
So I did profile this for the spatio temporal conditional moments

Constant time enhancement:
deterministic batching 179s, base 181s

Linear time enhancement:
deterministic batching 197s, base 198s

malihass:
I don't see a significant speed up when using it for the spatio temporal conditional moments. 

with deterministic batching I see 179s for 3 epochs, vs 181s without deterministic batching

bnb32:
> I don't see a significant speed up when using it for the spatio temporal conditional moments.
> 
> with deterministic batching I see 179s for 3 epochs, vs 181s without deterministic batching

Good to know that the original method doesn't have any unforeseen inefficiencies then. Thanks for checking. I'll probably just remove this branch.

PR #162, opened 2023-08-24
Bnb/h5 with era training

bnb32:


sup3r/bias/bias_calc.py
```diff
@@ -91,16 +101,22 @@ def __init__(self, base_fps, bias_fps, base_dset, bias_feature,
...
-                                         target=self.target, shape=self.shape,
-                                         val_split=0.0, **bias_handler_kwargs)
+        self.bias_dh = self.bias_handler(
+            self.bias_fps,
+            [self.bias_feature],
+            target=self.target,
+            shape=self.shape,
+            val_split=0.0,
+            **bias_handler_kwargs,
+        )
```
grantbuster, line 119: Did you run an auto formatter on everything? Do you like this style better and why? Some thoughts: The original formatting has strong horizontal separation of the output var, method, and input args. In the original formatting, it's very clear which lines are part of the function call. In the new formatting, it's more difficult to visually separate the output arg from the input args. Lastly, what is the logic for having the closing parenthesis on the 0 column? It's closing the function call, not the original line?

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -0,0 +1,614 @@
...
+
+    @property
+    def lr_time_index(self):
+        """Return time index for low_res data"""
+        return self._lr_time_index
+
+    @lr_time_index.setter
+    def lr_time_index(self, time_index):
+        """Set time index for low_res data"""
+        self._lr_time_index = time_index
```
grantbuster, line 326: There's no purpose in having a property and a setter if you're not including any additional logic or checks in either method. This is literally the same as exposing `self.lr_time_index` as a public unprotected attribute. You could argue it's nice to have the additional docstring but i think the variable names are self-explanatory and i dont think you can programmatically retrieve the docstring of a property/setter from a terminal anyway.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -0,0 +1,614 @@
...
+
+        self.hr_data = self.hr_dh.data[
+            : self.hr_required_shape[0],
+            : self.hr_required_shape[1],
+            : self.hr_required_shape[2],
+        ]
+        self.hr_time_index = self.hr_dh.time_index[: self.hr_required_shape[2]]
+        self.lr_time_index = self.lr_dh.time_index[: self.lr_required_shape[2]]
+
+    def _run_pair_checks(self, hr_handler, lr_handler):
```
grantbuster, line 215: maybe this is done somewhere else, but i'd recommend adding a check that the time index attributes match exactly.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -0,0 +1,614 @@
...
+            [
+                regridder(self.lr_input_data[..., i]).reshape(
+                    self.lr_required_shape
+                )[..., np.newaxis]
+                for i in range(len(self.features))
+            ],
+            axis=-1,
+        )
+
+    def get_next(self):
```
grantbuster, line 582: This method looks clean and well organized, nice job (except for the line break style that i'm not a fan of :))

sup3r/preprocessing/dual_batch_handling.py
```diff
@@ -0,0 +1,297 @@
...
+    def hr_sample_shape(self):
+        """Get sample shape for high_res data"""
+        return self.data_handlers[0].hr_dh.sample_shape
+
+    @property
+    def lr_sample_shape(self):
+        """Get sample shape for low_res data"""
+        return self.data_handlers[0].lr_dh.sample_shape
+
+    def __next__(self):
```
grantbuster, line 108: This method looks clean and well organized, nice job (except for the line break style that i'm not a fan of :))

grantbuster (changes requested):

sup3r/preprocessing/data_handling/dual_data_handling.py
bnb32: They shouldnt match exactly since this still allows for temporal enhancement but the shapes are checked. I could add a check that `hr_time_index[::t_enhance] = lr_time_index` though.

sup3r/preprocessing/data_handling/dual_data_handling.py
bnb32: Good point. These were cluttering things anyway.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -0,0 +1,614 @@
...
+
+        return np.concatenate(
+            [
+                regridder(self.lr_input_data[..., i]).reshape(
+                    self.lr_required_shape
+                )[..., np.newaxis]
+                for i in range(len(self.features))
+            ],
+            axis=-1,
+        )
```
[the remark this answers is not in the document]
bnb32: agreed

sup3r/preprocessing/dual_batch_handling.py
```diff
@@ -0,0 +1,297 @@
...
+                ].hr_val_data[val_index['hr_index']]
+                low_res[i, ...] = self.data_handlers[
+                    val_index['handler_index']
+                ].lr_val_data[val_index['lr_index']]
+                self._remaining_observations -= 1
+                self.current_batch_indices.append(val_index)
+
+            if self.sample_shape[2] == 1:
+                high_res = high_res[..., 0, :]
+                low_res = low_res[..., 0, :]
```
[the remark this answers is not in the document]
bnb32: yeah this was to check if we only have a single timestep e.g. spatial batch handler with 4D batches

sup3r/preprocessing/data_handling/dual_data_handling.py
grantbuster: I think that's worthwhile. The risk here is that for whatever reason the hr/lr datasets aren't of the same weather year. I suppose this might also be an issue for lat/lon although that might be more difficult to check.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -0,0 +1,475 @@
...
+            Array of regridded low_res data with all features
+            (spatial_1, spatial_2, temporal, n_features)
+        """
+        logger.info('Regridding low resolution feature data.')
+        regridder = self.get_regridder()
+
+        out = []
+        for i in range(len(self.features)):
+            tmp = regridder(self.lr_input_data[..., i]).reshape(
+                self.lr_required_shape)[..., np.newaixs]
```
grantbuster, line 444: better although i still think things look way more clean without line breaks. The real reason I'm writing this though is bc there's a typo: `np.newaixs`

grantbuster (changes requested):

sup3r/bias/bias_calc.py
grantbuster: @bnb32 it looks like some of the files are still black-formatted. Are you going to revert this or are we going with the black style?

grantbuster: I just think some formatting like in the screenshot below is total garbage (forward_pass.py). It looks like C++ nonsense and i think we're losing the readability of python. Some of this stuff is just the result of chaining too many commands into one "line". 

image

bnb32: Yeah I'm not a fan of that formatting either but I don't think it's that frequent and I personally consider the consistency a net positive. I think 
```out_arg = func(
   input_args
)``` is quite readable in general though.

Comment (grantbuster) [reply sup3r/bias/bias_calc.py]: I think it's a step backwards as far as code standards but if you want to do it for this repo i'll leave that decision to you. I think it's worst with massive chained "one-liners" like in the screenshot above. Basically i think this moves the onus to the developer to write smaller function chains to get more sensible code (e.g., split up the code above into 3-4 single lines that are actually single lines). IMO we should have better standards than in the example above. If we do want to do black across the board, i recommend adding the github action: https://github.com/marketplace/actions/run-black-formatter

Comment (bnb32) [reply sup3r/bias/bias_calc.py]: Do you think that onus is a bad one? I think the example above is the worst and for 99% the readability is maintained or improved tbh. I've turned off black formatting though since I don't want to enforce something you don't like, but I'd prefer not to have to manually revert a ton.

Comment (grantbuster) [reply sup3r/bias/bias_calc.py]: I don't think moving the focus onto a different type of style is bad. I do think black makes some code more readable and some other code much less so. I guess it just goes to show that these tools won't automatically transform any code into "good" code. There's still a lot of human considerations the developer has to work with to make readable code.

Comment (bnb32) [reply sup3r/bias/bias_calc.py]: Definitely agree!

File: sup3r/pipeline/forward_pass.py [hunk 9]
```diff
@@ -1446,60 +1405,35 @@ def pad_width(self):
...
-                    self.strategy.spatial_pad
-                    + s1_stop
-                    - self.strategy.grid_shape[0]
-                ),
-            )
-        )
+            np.maximum(0, (self.strategy.spatial_pad - s1_start)))
+        pad_s1_end = (self.strategy.spatial_pad + s1_stop
+                      - self.strategy.grid_shape[0])
+        pad_s1_end = int(np.maximum(0, pad_s1_end))
```
grantbuster, line 1419: Thank you! 🙌

grantbuster (approved):

sup3r/pipeline/forward_pass.py
bnb32: Sure thing!

PR #163, opened 2023-08-24
Gb/skill

grantbuster:


grantbuster:
@bnb32 can you think of any other common skill test metrics we should add? 

Here's an example of a high-complexity skill evaluation. Not sure how much value all of this adds... https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2022JD036659

and an example of a simple evaluation: https://www.sciencedirect.com/science/article/pii/S0048969721056576

I would really like to just present bias and KS test statistic, especially since we're going to be looking at more variables than other studies.

bnb32 (approved):
Straightforward. Looks good.

PR #165, opened 2023-09-11
Bnb/ws power law

bnb32:


bnb32:
@grantbuster Main edits in `feature_handling.py`, `interpolation.py`, and `test_data_handling_nc.py`

sup3r/preprocessing/feature_handling.py
```diff
@@ -724,6 +725,98 @@ def compute(cls, data, height):
         return rews
 
 
+class UWindPowerLaw(DerivedFeature):
+    """U wind component feature class with needed inputs method and compute
+    method. Uses power law extrapolation to get values above surface
+
+    https://en.wikipedia.org/wiki/Wind_profile_power_law
```
grantbuster, line 732: Thoughts on using this NOAA reference and set alpha=0.2 and put alpha as a class variable for transparency?

https://csl.noaa.gov/projects/lamar/windshearformula.html

tests/data/test_era5_co_2012.nc
grantbuster, line 1: Just a reminder to be really intentional when adding big test files like this. Every file added is added forever and bloats the repo size.

tests/data_handling/test_data_handling_nc_cc.py
```diff
@@ -1,22 +1,42 @@
...
+    DataHandlerNCforCCwithPowerLaw,
+)
+
+
+def test_data_handling_nc_cc_power_law():
+    """Make sure the power law extrapolation of wind operates correctly"""
+    input_files = [os.path.join(TEST_DATA_DIR, 'uas_test.nc')]
+
+    with xr.open_mfdataset(input_files) as fh:
+        u_100m = fh['uas'].values * 100**0.143
```
grantbuster, line 22: if you set alpha to class attr you can reference that here

grantbuster (changes requested):
minor thought on power law reference and alpha value

sup3r/preprocessing/feature_handling.py
bnb32: Class var is a good idea. Why the disagreement between the alpha values though?

tests/data/test_era5_co_2012.nc
bnb32: Yeah this is trimmed for only needed vars but I misnamed it last merge.

sup3r/preprocessing/feature_handling.py
grantbuster: I'm not sure but NOAA is more defensible haha

grantbuster: our high school teachers would be so proud

grantbuster (approved):

sup3r/preprocessing/feature_handling.py
bnb32: Our high school teachers were wrong. Wikipedia actually links to a paper. NOAA doesn't have a source for 0.2 that I can see?

PR #169, opened 2023-10-04
Bnb/exo refactor plus

bnb32:


sup3r/models/abstract.py
```diff
@@ -274,13 +268,10 @@ def _combine_fwp_output(self, hi_res, exogenous_data=None):
             (n_obs, spatial_1, spatial_2, n_features)
             (n_obs, spatial_1, spatial_2, n_temporal, n_features)
         exogenous_data : dict | None
```
grantbuster, line 270: Update class spec

sup3r/models/abstract.py
```diff
@@ -1309,6 +1260,10 @@ def generate(self,
             (n_obs, spatial_1, spatial_2, n_features)
             (n_obs, spatial_1, spatial_2, n_temporal, n_features)
         """
+        if (isinstance(exogenous_data, dict)
+                and not isinstance(exogenous_data, ExoData)):
+            exogenous_data = ExoData(exogenous_data)
```
grantbuster, line 1265: I don't think you need the double `isinstance()` call do you? Or is this just because ExoData inherits dict? Either way, just use the more restrictive of the two conditions. 

Also, since this is just two lines, why not just put this everywhere in this file that accepts exogenous_data input so it can always be dict/exodata/none

sup3r/models/multi_step.py
```diff
@@ -299,133 +258,6 @@ def model_params(self):
         return tuple(model.model_params for model in self.models)
 
 
-class SpatialThenTemporalBase(MultiStepGan):
```
grantbuster, line 302: great!

sup3r/preprocessing/data_handling/exogenous_data_handling.py
```diff
@@ -19,6 +19,172 @@
...
+    """Special dictionary class for multiple exogenous_data steps"""
+
+    def __init__(self, steps):
+        """Combine multiple SingleExoDataStep objects
+
+        Parameters
+        ----------
+        steps : list | dict
+            List of SingleExoDataStep objects or a feature dictionary with list
+            of steps for each feature
```
grantbuster, line 65: I think this is the place to be a lot more descriptive and give an example. Not immediately obvious to me how it could be either a list or a dict. Do we even need to handle both? Python is flexibly typed but that doesn't mean it's always a good idea to accept multiple types. Often much cleaner and more explicit if we just lean into a single type spec. 

Also, I suggest giving an example like this (copied from the previous docstrings): 

```
            Dictionary of exogenous feature data with entries describing
            whether features should be combined at input, a mid network layer,
            or with output. This doesn't have to include the 'model' key since
            this data is for a single step model. e.g.
            {'topography': {'steps': [
                {'combine_type': 'input', 'data': ..., 'resolution': ...},
                {'combine_type': 'layer', 'data': ..., 'resolution': ...}]}}

File: sup3r/preprocessing/data_handling/exogenous_data_handling.py [hunk 5]
```diff
@@ -19,6 +19,172 @@
...
+        only model steps greater than or equal to split_step.
+
+        Parameters
+        ----------
+        split_step : int
+            Step index to use for splitting. To split this into exo data for
+            spatial models and temporal models split_step should be
+            len(spatial_models). If this is for a TemporalThenSpatial model
+            split_step should be len(temporal_models).
+        exogenous_data : dict
```
grantbuster, line 120: not an input?

grantbuster (changes requested):

grantbuster (approved):

PR #173, opened 2023-11-03
Bnb/dev

bnb32:
Bc code for using vortex vs era means to compute factors. Some small tweaks for dual handler use with different sets of features for lr and hr handlers.


sup3r/bias/bias_correct_means.py
```diff
@@ -0,0 +1,949 @@
...
+
+
+class VortexMeanPrepper:
+    """Class for converting monthly vortex tif files for each height to a
+    single h5 files containing all monthly means for all requested output
+    heights.
+    """
+
+    def __init__(self, path_pattern, in_heights, out_heights, overwrite=False):
+        """Parameters
```
grantbuster, line 35: Minor but i wonder if you need a line break to get the parameters to show up in the docs page

sup3r/bias/bias_correct_means.py
```diff
@@ -0,0 +1,949 @@
+"""Classes to compute means from vortex and era data and compute bias
+correction factors.
```
grantbuster, line 2: Do you want to include any details on where / how to download the vortex means? No code just a quick tip.

sup3r/bias/bias_correct_means.py
```diff
@@ -0,0 +1,949 @@
...
+    """Class for converting monthly vortex tif files for each height to a
+    single h5 files containing all monthly means for all requested output
+    heights.
+    """
+
+    def __init__(self, path_pattern, in_heights, out_heights, overwrite=False):
+        """Parameters
+        ----------
+        path_pattern : str
+            Pattern for input tif files. Needs to include {month} and {height}
```
grantbuster, line 38: surprised no assert statement in the init for these format keys

sup3r/bias/bias_correct_means.py
```diff
@@ -0,0 +1,949 @@
...
+            elif os.path.exists(fp_out):
+                logger.info(f"{fp_out} already exists and overwrite=False.")
+
+    @classmethod
+    def run(
+        cls, path_pattern, in_heights, out_heights, fp_out, overwrite=False
+    ):
+        """Read vortex tif files, convert these to monthly netcdf files for all
+        input heights, interpolate this data to requested output heights, mask
+        fill values, and write all data to h5 file.
```
grantbuster, line 306: Seems kind of weird that you need the netcdf intermediate step but im assuming you thought of this and need it

sup3r/bias/bias_correct_means.py
```diff
@@ -0,0 +1,949 @@
...
+        }
+        df = pd.DataFrame.from_dict(out)
+        df["latitude"] = self.lats.flatten()
+        df["longitude"] = self.lons.flatten()
+        df["gid"] = np.arange(len(df["latitude"]))
+        df.to_csv(out_file)
+        logger.info(f"Finished writing means for {out_file}.")
+
+    @classmethod
+    def run(cls, era_pattern, years, features, out_pattern):
```
grantbuster, line 457: The ERA netcdf files should be able to be handled by the standard DataHandlerNC classes and the BiasCalc class. Why didnt you just convert vortex data to a standard format and then use the BiasCalc class from there? I think the mods to allow the BiasCalc module to use monthly data would be really simple

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -162,24 +162,27 @@ def normalize(self, means, stdevs):
...
+                        means=means[indices],
+                        stds=stdevs[indices],
                         max_workers=self.hr_dh.norm_workers)
 
     @property
     def output_features(self):
         """Get list of output features. e.g. those that are returned by a
         GAN"""
-        return self.hr_dh.output_features
+        return self.lr_dh.output_features
```
grantbuster, line 185: This is surprising, maybe i'm missing something, why wouldnt the high-res data handler have the correct output feature list? Just because the low-res handler has a list of "train only" features that makes this correct?

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -219,7 +222,7 @@ def _run_pair_checks(self, hr_handler, lr_handler):
         assert hr_handler.val_split == 0 and lr_handler.val_split == 0, msg
         msg = ('Handlers have incompatible number of features. '
                f'({hr_handler.features} vs {lr_handler.features})')
-        assert hr_handler.features == lr_handler.features, msg
+        assert hr_handler.features == self.output_features, msg
```
grantbuster, line 225: Oh maybe this confirms my previous question?

grantbuster (changes requested):

grantbuster:
@bnb32 my only questions is whether we want to keep the `BiasCorrectionFromMeans` or are you integrating with the `bias_calc.py` module? Also, i guess scalar bias correction is linear but i'd still recommend a shorter name than `MonthlyLinearCorrectionScalarOnly`, maybe `MonthlyScalarCorrection`

bnb32:
> @bnb32 my only questions is whether we want to keep the `BiasCorrectionFromMeans` or are you integrating with the `bias_calc.py` module? Also, i guess scalar bias correction is linear but i'd still recommend a shorter name than `MonthlyLinearCorrectionScalarOnly`, maybe `MonthlyScalarCorrection`

Are you asking whether we should move the code into bias_calc.py or whether the BiasCorrectionFromMeans class should be removed due to redundant code? This class is for bias correcting existing h5 files so I dont think the latter makes sense. Maybe it does make sense to move code though.

Ill shorten the name, good suggestion.

grantbuster (approved):

PR #174, opened 2023-11-04
Add ability to specify duplicate command names in Pipeline

ppinchuk:
Short example on how to specify duplicate pipeline steps is [here](https://nrel.github.io/gaps/misc/examples.users.html#duplicate-pipeline-steps).

### DIFF
--- a/sup3r/bias/bias_calc.py
+++ b/sup3r/bias/bias_calc.py
@@ -211,7 +211,8 @@ class DataRetrievalBase:
                f"{fun_str};\n"
                "t_elap = time.time() - t0;\n")
 
-        cmd = BaseCLI.add_status_cmd(config, ModuleName.BIAS_CALC, cmd)
+        pipeline_step = config.get('pipeline_step') or ModuleName.BIAS_CALC
+        cmd = BaseCLI.add_status_cmd(config, pipeline_step, cmd)
         cmd += ";\'\n"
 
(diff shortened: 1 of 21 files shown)


bnb32 (commented):
Looks great, man! Seems like this was just a bit tedious but easier than expected? Based on the test it looks like pipeline_step doesn't need to be specified explicitly, but I'm not seeing where in the code the implicit specification is handled?

ppinchuk:
Yea, you're 100% right about it being easier than expected. When I was ding the GAPs refactor, I realized it would be a non-negligible amount of changes to integrate this feature, which I had no idea if you guys needed or even wanted. I think at that time I decided to take the conservative approach and told myself "sup3r will just behave as if its still running on the reV pipeline". I made a mental note of that and moved on, which is probably why I was remembering that it would be "difficult". Now that there's appetite for this, I think the 100 extra lines are totally worth it.

The [`pipeline_step` parsing is handled by GAPs](https://github.com/NREL/gaps/blob/080f0fc8f1462664a3e4310d907db8b990d5e8d2/gaps/pipeline.py#L26-L57), which then passes it as an argument to the `from_config` functions in sup3r. The parsing step is super basic, so let me know if you have any better ideas for how to specify these in the config and read them into the code.

bnb32 (approved):

PR #177, opened 2023-11-11
stdev and mean hot fix

bnb32:
Previously computation of multi handler mean and stdev implicitly assumed each handler had roughly the same number of elements. Changed this to include handler weights.

### DIFF
--- a/sup3r/models/abstract.py
+++ b/sup3r/models/abstract.py
@@ -858,10 +858,12 @@ class AbstractSingleModel(ABC):
 
         Parameters
         ----------
-        loss : str
+        loss : str | dict
             Loss function class name from sup3r.utilities.loss_metrics
             (prioritized) or tensorflow.keras.losses. Defaults to
-            tf.keras.losses.MeanSquaredError.
+            tf.keras.losses.MeanSquaredError. This can be provided as a dict
+            with kwargs for loss functions with extra parameters.
+            e.g. {'SpatialExtremesLoss': {'weight': 0.5}}
 
         Returns
         -------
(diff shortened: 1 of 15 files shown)


sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -19,7 +19,15 @@
 # pylint: disable=unsubscriptable-object
 class DualDataHandler(CacheHandlingMixIn, TrainingPrepMixIn):
     """Batch handling class for h5 data as high res (usually WTK) and netcdf
-    data as low res (usually ERA5)"""
+    data as low res (usually ERA5)
+
+    NOTE: When initializing the lr_handler it's important to pick a shape
```
grantbuster, line 24: Nice, i'd like to see us add more of these "gotchas" directly to the docstrings

grantbuster (approved):

sup3r/preprocessing/data_handling/dual_data_handling.py
bnb32: agreed!

sup3r/models/abstract.py
```diff
@@ -870,6 +872,9 @@ def get_loss_fun(loss):
             "MeanSquaredError" is requested, this will return
             an instance of tf.keras.losses.MeanSquaredError()
         """
+        kwargs = {}
+        if isinstance(loss, dict):
+            loss, kwargs = next(iter(loss.items()))
```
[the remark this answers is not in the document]
bnb32: Yeah this was based on a helpful hint from ruff. Itll only return the first key/value even if there are multiple entries.

PR #179, opened 2023-11-22
bug fix with negative test for dual DH operating on data handlers tha…

grantbuster:
…t did not load from cache

### DIFF
--- a/sup3r/preprocessing/data_handling/dual_data_handling.py
+++ b/sup3r/preprocessing/data_handling/dual_data_handling.py
@@ -176,6 +176,14 @@ class DualDataHandler(CacheHandlingMixIn, TrainingPrepMixIn):
         -------
         dict
         """
+
+        if self.lr_dh.data is None or self.hr_dh.data is None:
+            msg = ('Low-res or High-res DataHandler object has '
+                   'DataHandler.data=None! Try initializing DualDataHandler '
+                   'with load_cached=True')
+            logger.error(msg)
+            raise RuntimeError(msg)
+
         out = copy.deepcopy(self.hr_dh.means)
         out.update(self.lr_dh.means)
         return out
@@ -190,6 +198,14 @@ class DualDataHandler(CacheHandlingMixIn, TrainingPrepMixIn):
         -------
         dict
         """
+
+        if self.lr_dh.data is None or self.hr_dh.data is None:
+            msg = ('Low-res or High-res DataHandler object has '
+                   'DataHandler.data=None! Try initializing DualDataHandler '
+                   'with load_cached=True')
+            logger.error(msg)
+            raise RuntimeError(msg)
+
         out = copy.deepcopy(self.hr_dh.stds)
         out.update(self.lr_dh.stds)
         return out
@@ -213,10 +229,19 @@ class DualDataHandler(CacheHandlingMixIn, TrainingPrepMixIn):
             Max workers to perform normalization. if None, self.norm_workers
             will be used
         """
+
+        if self.lr_dh.data is None or self.hr_dh.data is None:
+            msg = ('Low-res or High-res DataHandler object has '
+                   'DataHandler.data=None! Try initializing DualDataHandler '
+                   'with load_cached=True')
+            logger.error(msg)
+            raise RuntimeError(msg)
+
         if means is None:
             means = self.means
         if stds is None:
             stds = self.stds
+
         logger.info('Normalizing low resolution data features='
                     f'{self.lr_dh.features}')
         self.lr_dh.normalize(means=means, stds=stds, max_workers=max_workers)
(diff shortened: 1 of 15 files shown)


bnb32 (changes requested):
I think instead of requiring the `lr_handler` cache to be loaded we could move the stats methods into the `TrainingPrepMixIn` class so they're inherited by dual handler and then have the methods operate directly on `lr_data`. 

For example, the `DualDataHandler._get_stats` method could be:

```
super()._get_stats()
self.hr_dh._get_stats()
```

where `super()._get_stats()` will define `self._means` and `self._stds`.

and the normalize method could be

```
super().normalize(...)
self.hr_dh.normalize(...)
```

The `.means` property could be

```
out = copy.deepcopy(self.hr_dh.means)
out.update(self._means)
return out
```

etc.

Maybe I'm missing it but I don't see `lr_data` data being normalized, just `lr_handler.data`, but `lr_data` is what is sampled by the `get_next` method.

grantbuster:
@bnb32 GO ENJOY YOUR DAY OFF

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -253,18 +251,15 @@ def normalize(self, means=None, stds=None, max_workers=None):
         super().normalize(means=means, stds=stds,
                           features=self.lr_dh.features,
                           max_workers=self.lr_dh.norm_workers)
+        self.lr_dh.normalize(means=means, stds=stds,
```
bnb32, line 254: Why do we need this? We should be able to compute stats without loading lr_dh.data.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -183,10 +183,9 @@ def means(self):
         dict
         """
 
-        if self.lr_dh.data is None or self.hr_dh.data is None:
-            msg = ('Low-res or High-res DataHandler object has '
-                   'DataHandler.data=None! Try initializing DualDataHandler '
-                   'with load_cached=True')
+        if self.hr_dh.data is None:
+            msg = ('High-res DataHandler object has DataHandler.data=None! '
```
bnb32, line 187: Maybe this should say try initializing the high-res handler with load_cached=True.

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -113,7 +113,7 @@ def get_data(self):
         """Check hr and lr shapes and trim hr data if needed to match required
         relationship to lr shape based on enhancement factors. Then regrid lr
         data and split hr and lr data into training and validation sets."""
-        self._shape_check()
+        self._set_hr_data()
```
bnb32, line 116: Much better name

bnb32 (changes requested):

sup3r/preprocessing/data_handling/dual_data_handling.py
grantbuster: updated

sup3r/preprocessing/data_handling/dual_data_handling.py
grantbuster: this confused me so much hahaha

sup3r/preprocessing/data_handling/dual_data_handling.py
bnb32: Ok but you're forcing it to be loaded a few lines above.

grantbuster: I just fixed that exception raise clause in the last commit (i think this is what you're talking about and i agree this was leftover from previous)

bnb32: Ok nice. Works for me!

sup3r/preprocessing/data_handling/dual_data_handling.py
bnb32: Lol yeah it's a confusing method for sure

sup3r/preprocessing/data_handling/dual_data_handling.py
```diff
@@ -253,18 +251,15 @@ def normalize(self, means=None, stds=None, max_workers=None):
...
                              features=self.hr_dh.features,
                              max_workers=self.hr_dh.norm_workers)
 
-        # need to normalize data attribute arrays in addition to handlers
-        mean_arr = np.array([means[fn] for fn in self.lr_dh.features])
-        std_arr = np.array([stds[fn] for fn in self.lr_dh.features])
-        self.lr_data = (self.lr_data - mean_arr) / std_arr
-        self.lr_data = self.lr_data.astype(np.float32)
-
         if id(self.hr_data.base) != id(self.hr_dh.data):
```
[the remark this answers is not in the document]
bnb32: Well I think it would just get updated after `val_data` is set since the `val_split` method just updates the `hr_dh.data` attribute, but I agree it might be confusing. Theres also the option of just using the `_normalize` method to directory normalize hr_data but since you want everything normalized maybe this current way makes the most sense.

bnb32 (approved):

sup3r/preprocessing/data_handling/dual_data_handling.py
grantbuster: Honestly just the name was confusing, I kept trying to figure out where hr_data was being set and took me a while to realize this method did that

PR #180, opened 2023-11-27
Bnb/ddh norm

bnb32:


sup3r/preprocessing/batch_handling.py
```diff
@@ -618,14 +618,15 @@ def _parallel_normalization(self):
...
+                dh.normalize(self.means, self.stds,
+                             max_workers=dh.norm_workers)
         else:
             with ThreadPoolExecutor(max_workers=max_workers) as exe:
                 futures = {}
                 now = dt.now()
                 for idh, dh in enumerate(self.data_handlers):
                     future = exe.submit(dh.normalize, self.means, self.stds,
-                                        max_workers=1)
+                                        max_workers=dh.norm_workers)
```
grantbuster, line 629: Is there a specific reason you would want a worker to spin up more threads within a thread pool? Historically i've been very cautious of this. I don't want my threads/processes to be forking themselves.

sup3r/preprocessing/data_handling/base.py
```diff
@@ -1269,14 +1200,12 @@ def serial_data_fill(self, shifted_time_chunks):
...
-            if t % interval == 0:
-                logger.info(f'Added {t + 1} of {len(shifted_time_chunks)} '
-                            'chunks to final data array')
+                self._data_fill(t, ts, f_index, f)
+            logger.info(f'Added {t + 1} of {len(shifted_time_chunks)} '
+                        'chunks to final data array')
             self._raw_data.pop(t)
 
-    def parallel_data_fill(self, shifted_time_chunks, max_workers=None):
+    def data_fill(self, shifted_time_chunks, max_workers=None):
```
grantbuster, line 1208: Not a huge fan of this naming convention. I always find it confusing when there is a `.method` and a `._method`. I think you're trying to hide the serial method from the "public" view but it makes the developer's job more confusing. I'd suggest `_single_data_fill()` and `_data_fill()`

grantbuster (changes requested):
Might have a few negative tests fail that i added recently, i can handle them if you want.

sup3r/preprocessing/batch_handling.py
bnb32: Yeah that's fair. Makes sense to go back to 1

grantbuster (approved):

PR #184, opened 2024-02-07
added kwarg to fwp to allow for constant tensor output e.g., in the c…

grantbuster:
…ase of precip=0

### DIFF
--- a/sup3r/pipeline/forward_pass.py
+++ b/sup3r/pipeline/forward_pass.py
@@ -610,7 +610,8 @@ class ForwardPassStrategy(InputMixIn, DistributedProcess):
                  exo_kwargs=None,
                  bias_correct_method=None,
                  bias_correct_kwargs=None,
-                 max_nodes=None):
+                 max_nodes=None,
+                 allowed_const=False):
         """Use these inputs to initialize data handlers on different nodes and
         to define the size of the data chunks that will be passed through the
         generator.
@@ -715,6 +716,15 @@ class ForwardPassStrategy(InputMixIn, DistributedProcess):
         max_nodes : int | None
             Maximum number of nodes to distribute spatiotemporal chunks across.
             If None then a node will be used for each temporal chunk.
+        allowed_const : list | bool
+            Tensorflow has a tensor memory limit of 2GB (result of protobuf
+            limitation) and when exceeded can return a tensor with a
+            constant output. sup3r will raise a ``MemoryError`` in response. If
+            your model is allowed to output a constant output, set this to True
+            to allow any constant output or a list of allowed possible constant
+            outputs. For example, a precipitation model should be allowed to
+            output all zeros so set this to ``[0]``. For details on this limit:
+            https://github.com/tensorflow/tensorflow/issues/51870
         """
         self._input_handler_kwargs = input_handler_kwargs or {}
         target = self._input_handler_kwargs.get('target', None)
@@ -749,6 +759,7 @@ class ForwardPassStrategy(InputMixIn, DistributedProcess):
         self._lr_lat_lon = None
         self._init_handler = None
         self._handle_features = None
+        self.allowed_const = allowed_const
 
         self._single_ts_files = self._input_handler_kwargs.get(
             'single_ts_files', None)
@@ -1779,9 +1790,20 @@ class ForwardPass:
         out_data : ndarray
             Forward pass output corresponding to the given chunk index
         """
+
+        allowed_const = self.strategy.allowed_const
+        if allowed_const is True:
+            return
+        elif allowed_const is False:
+            allowed_const = []
+        elif not isinstance(allowed_const, (list, tuple)):
+            allowed_const = [allowed_const]
+
         for i, f in enumerate(self.output_features):
             msg = f'All spatiotemporal values are the same for {f} output!'
-            if np.all(out_data[0, 0, 0, i] == out_data[..., i]):
+            value0 = out_data[0, 0, 0, i]
+            all_same = (value0 == out_data[..., i]).all()
+            if all_same and value0 not in allowed_const:
                 self.strategy.failed_chunks = True
                 logger.error(msg)
                 raise MemoryError(msg)


bnb32 (changes requested):
Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?

grantbuster:
> Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?

No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.

bnb32:
> > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> 
> No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.

Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?

grantbuster:
> > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > 
> > 
> > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> 
> Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?

Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?

bnb32:
> > > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > > 
> > > 
> > > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> > 
> > 
> > Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?
> 
> Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?

I think if we for sure know its a 2GB cap then the second option wouldnt be too bad. We could just do `mem = (lats * lons * times * n_obs * s_enhance**2 * t_enhance * n_filters * dtype_size)` right? And this could just be a base model attribute. But yeah if we don't know for sure whats causing this then I guess checking constant output is still the way to go. I don't love the first option either, prob bc it feels kind of hacky.

grantbuster:
> > > > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > > > 
> > > > 
> > > > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> > > 
> > > 
> > > Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?
> > 
> > 
> > Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?
> 
> I think if we for sure know its a 2GB cap then the second option wouldnt be too bad. We could just do `mem = (lats * lons * times * n_obs * s_enhance**2 * t_enhance * n_filters * dtype_size)` right? And this could just be a base model attribute. But yeah if we don't know for sure whats causing this then I guess checking constant output is still the way to go. I don't love the first option either, prob bc it feels kind of hacky.

but then you have to go scrape the model for the enhancement factors and the biggest combination of enhancement * filter count... And this is only considering the convolutional layers, what if we introduce new custom layers with weirder things going on? Like the concatenation layers will add a new filter channel but it's hard to parse that out.

bnb32:
> > > > > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > > > > 
> > > > > 
> > > > > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> > > > 
> > > > 
> > > > Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?
> > > 
> > > 
> > > Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?
> > 
> > 
> > I think if we for sure know its a 2GB cap then the second option wouldnt be too bad. We could just do `mem = (lats * lons * times * n_obs * s_enhance**2 * t_enhance * n_filters * dtype_size)` right? And this could just be a base model attribute. But yeah if we don't know for sure whats causing this then I guess checking constant output is still the way to go. I don't love the first option either, prob bc it feels kind of hacky.
> 
> but then you have to go scrape the model for the enhancement factors and the biggest combination of enhancement * filter count... And this is only considering the convolutional layers, what if we introduce new custom layers with weirder things going on? Like the concatenation layers will add a new filter channel but it's hard to parse that out.

Well the models are already scraped for enhancement factors, so we're partially there. https://github.com/NREL/sup3r/blob/92ac0310cd4556aee96d0d99ed3db75dcbcbae0e/sup3r/models/abstract.py#L110

grantbuster:
> > > > > > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > > > > > 
> > > > > > 
> > > > > > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> > > > > 
> > > > > 
> > > > > Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?
> > > > 
> > > > 
> > > > Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?
> > > 
> > > 
> > > I think if we for sure know its a 2GB cap then the second option wouldnt be too bad. We could just do `mem = (lats * lons * times * n_obs * s_enhance**2 * t_enhance * n_filters * dtype_size)` right? And this could just be a base model attribute. But yeah if we don't know for sure whats causing this then I guess checking constant output is still the way to go. I don't love the first option either, prob bc it feels kind of hacky.
> > 
> > 
> > but then you have to go scrape the model for the enhancement factors and the biggest combination of enhancement * filter count... And this is only considering the convolutional layers, what if we introduce new custom layers with weirder things going on? Like the concatenation layers will add a new filter channel but it's hard to parse that out.
> 
> Well the models are already scraped for enhancement factors, so we're partially there.
> 
> https://github.com/NREL/sup3r/blob/92ac0310cd4556aee96d0d99ed3db75dcbcbae0e/sup3r/models/abstract.py#L110

Yeah yeah but the rest is challenging, you need to check every layer for what size tensor it would output. You could have a bigger tensor prior to s/t enhancement because of a ton of filters or something else going on because of funky custom layers like the high-res concat layers. I think the current approach is fine.

bnb32:
> > > > > > > > Haha. Nice new wrench in the annoying saga of constant output. Can we just return a memory error if the output tensor exceeds 2GB instead of doing the constant output check?
> > > > > > > 
> > > > > > > 
> > > > > > > No haha we've talked about this, the problem is that the tensors mid-network are memory hogs thanks to their filter channels and we don't inspect this mid-network. I guess we could inspect this mid network but seems like a pain.
> > > > > > 
> > > > > > 
> > > > > > Oh right. I guess the easiest way to inspect mid network would be to add a mem check to the inference loop through the layers. We could also just check the model layers for max number of filter channels and use this multiplier, and array input size, to check max memory?
> > > > > 
> > > > > 
> > > > > Yeah, the arithmetic in the second option sounds complicated and you would have to know information about the s/t enhancement. Adding a memory check in the inference loop sounds like an option but i don't loveeee it... Hard to explain why. I guess we just don't really understand exactly whats going on here and checking bad outputs seems like the most sure-fire way of doing this?
> > > > 
> > > > 
> > > > I think if we for sure know its a 2GB cap then the second option wouldnt be too bad. We could just do `mem = (lats * lons * times * n_obs * s_enhance**2 * t_enhance * n_filters * dtype_size)` right? And this could just be a base model attribute. But yeah if we don't know for sure whats causing this then I guess checking constant output is still the way to go. I don't love the first option either, prob bc it feels kind of hacky.
> > > 
> > > 
> > > but then you have to go scrape the model for the enhancement factors and the biggest combination of enhancement * filter count... And this is only considering the convolutional layers, what if we introduce new custom layers with weirder things going on? Like the concatenation layers will add a new filter channel but it's hard to parse that out.
> > 
> > 
> > Well the models are already scraped for enhancement factors, so we're partially there.
> > https://github.com/NREL/sup3r/blob/92ac0310cd4556aee96d0d99ed3db75dcbcbae0e/sup3r/models/abstract.py#L110
> 
> Yeah yeah but the rest is challenging, you need to check every layer for what size tensor it would output. You could have a bigger tensor prior to s/t enhancement because of a ton of filters or something else going on because of funky custom layers like the high-res concat layers. I think the current approach is fine.

Yeah that's fair.

bnb32 (approved):

PR #186, opened 2024-02-14
Bnb/dev

bnb32:
While we're adding loss functions...


sup3r/utilities/era_downloader.py
```diff
@@ -622,53 +624,56 @@ def all_months_exist(cls, year, file_pattern):
...
-                            var, old_var.dtype, dimensions=old_var.dimensions)
-                        new_ds[var][:] = vals
-                        if hasattr(old_var, 'units'):
-                            new_ds[var].units = old_var.units
-                        if hasattr(old_var, 'standard_name'):
-                            standard_name = old_var.standard_name
-                            new_ds[var].standard_name = standard_name
-                        if hasattr(old_var, 'long_name'):
-                            new_ds[var].long_name = old_var.long_name
+                for var in keep_vars:
```
grantbuster, line 663: I like this semantic naming much better than working with unnamed list comp's. Nice!

sup3r/preprocessing/data_handling/base.py
```diff
@@ -1421,358 +1418,3 @@ def get_next(self, temporal_weights=None, spatial_weights=None):
             temporal_weights=temporal_weights, spatial_weights=spatial_weights)
         observation = self.data[self.current_obs_index]
         return observation
-
-    @classmethod
-    def get_file_times(cls, file_paths, **kwargs):
```
grantbuster, line 1426: Where did all these methods go?? I dont see them moved in this PR?

sup3r/utilities/loss_metrics.py
```diff
@@ -117,6 +117,103 @@ def __call__(self, x1, x2, sigma=1.0):
...
+        -------
+        tf.tensor
+            0D tensor with loss value
+        """
+        hub_heights = x1.shape[-1] // 2
+
+        x1_div = tf.stack(
+            [self._compute_md(x1, fidx=i) for i in range(2 * hub_heights)])
+        x2_div = tf.stack(
+            [self._compute_md(x2, fidx=i) for i in range(2 * hub_heights)])
```
grantbuster, line 209: Shouldn't this be `range(0, 2*hh, 2)`? Right now you're iterating from 0 to 2*hh and when i=1 I think that `self._compute_md()` will use i=(1,2) as u10m/v10m but i=(1,2) is really v10m/u100m

Actually i don't think thats quite right because `uidx = 2 * (fidx // 2)` but still, won't you be double calculating if you do `i in range(2*hh)`?

sup3r/utilities/loss_metrics.py
```diff
@@ -117,6 +117,103 @@ def __call__(self, x1, x2, sigma=1.0):
...
+        features is either 2 * N or 2 * N + 1
+
+        Parameters
+        ----------
+        x1 : tf.tensor
+            synthetic generator output
+            (n_observations, spatial_1, spatial_2, temporal, features)
+        x2 : tf.tensor
+            high resolution data
+            (n_observations, spatial_1, spatial_2, temporal, features)
```
grantbuster, line 197: somewhere you should `assert len(shape) == 5`

sup3r/utilities/loss_metrics.py
```diff
@@ -117,6 +117,103 @@ def __call__(self, x1, x2, sigma=1.0):
...
 
 
+class MaterialDerivativeLoss(tf.keras.losses.Loss):
+    """Loss class for the material derivative. This is the left hand side of
+    the Navier-Stokes equation and is equal to internal + external forces
+    divided by density.
+    https://en.wikipedia.org/wiki/Material_derivative
+    """
+
+    MAE_LOSS = MeanAbsoluteError()
```
grantbuster, line 127: For some of these loss functions we're still assuming MAE is the go-to metric but it is not always. Consider making the class attribute simply `LOSS_METRIC` or `ERROR_METRIC`. Personally I'm 100% fine with manipulating the class attributes at runtime but the name is presumptious.

sup3r/utilities/loss_metrics.py
```diff
@@ -117,6 +117,103 @@ def __call__(self, x1, x2, sigma=1.0):
...
+                              (x[:, 2:] - x[:, :-2]) / 2,
+                              x[:, -1:] - x[:, -2:-1]], axis=1)
+        elif axis == 2:
+            return tf.concat([x[..., 1:2, :] - x[..., 0:1, :],
+                              (x[..., 2:, :] - x[..., :-2, :]) / 2,
+                              x[..., -1:, :] - x[..., -2:-1, :]], axis=2)
+        elif axis == 3:
+            return tf.concat([x[..., 1:2] - x[..., 0:1],
+                              (x[..., 2:] - x[..., :-2]) / 2,
+                              x[..., -1:] - x[..., -2:-1]], axis=3)
```
grantbuster, line 150: ignorant question: why are you doing a `[2:]` offset to calculate the derivative instead of just doing a 1-pixel difference?

sup3r/utilities/loss_metrics.py
```diff
@@ -117,6 +117,103 @@ def __call__(self, x1, x2, sigma=1.0):
         return mmd
 
 
+class MaterialDerivativeLoss(tf.keras.losses.Loss):
```
grantbuster, line 120: you should add a test for this

grantbuster (changes requested):

sup3r/utilities/loss_metrics.py
grantbuster: Gotcha that makes sense, maybe consider just adding this note for posterity?

grantbuster (approved):

sup3r/utilities/loss_metrics.py
bnb32: Yeah good idea

PR #190, opened 2024-02-28
Defining some minimum versions

castelao:
These are the versions that I could install with Python-3.8. This info might be helpful to keep track.

Can you confirm that these versions don't conflict with your local & HPC environments?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,10 +5,10 @@ NREL-gaps>=0.6.0
 NREL-farms>=1.0.4
 google-auth-oauthlib==0.5.3
 pytest>=5.2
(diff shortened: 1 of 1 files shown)


bnb32:
> These are the versions that I could install with Python-3.8. This info might be helpful to keep track.
> 
> Can you confirm that these versions don't conflict with your local & HPC environments?

These all work for me.

castelao:
Thanks @bnb32 ! Let's hold this PR for a few days and give time for everyone to check it.

castelao:
@bnb32 @grantbuster , is it fine to move this one forward?

PR #191, opened 2024-02-29
Caching PIP

castelao:
Cache PIP to speed up a little the testing job.

Also, update actions to avoid warning messages.


bnb32 (approved):
Loving all this clean up you're doing. Thanks!

PR #196, opened 2024-03-14
Syntax update using numpy and scipy

castelao:
Minor editing that would stop some warnings about a future change in syntax.

I also added explicit dependency on `numpy` and `scipy` since those are explicitly called.


bnb32 (approved):
Looks fine to me. Not sure what's going on with those tests though. 

Apparently tensorflow 2.16 made some big API changes. I added <2.16 to the requirements in my PR.

PR #199, opened 2024-03-16
Sup3rWind examples 

bnb32:
Basically the same as Sup3rCC examples.


sup3r/postprocessing/file_handling.py
```diff
@@ -595,9 +587,40 @@ def _write_output(cls, data, features, lat_lon, times, out_file,
             attrs = {k: v if isinstance(v, str) else json.dumps(v)
                      for k, v in meta_data.items()}
 
-        with xr.Dataset(data_vars=data_vars, coords=coords,
-                        attrs=attrs) as ncfile:
-            ncfile.to_netcdf(out_file)
+        return xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
```
castelao, line 590: I like to add some metadata that might be helpful later. For instance, created/modified datetime might help to decide between two equivalent datasets.
```
attrs['date_modified'] = datetime.utcnow().isoformat()
if 'date_created' not in attrs:
    attrs['date_created'] = attrs['date_modified']
ds = xr.Dataset(...)
...
```

castelao (commented):
I don't have much to add here. I think it was a good move to create `_get_xr_dset()` thus allowing a pipeline directly from an `xr.Dataset` without necessarily creating the netCDF.

bnb32:
Merging this since it is low stakes and referenced in slides that are about to be published.

PR #202, opened 2024-03-21
Fix documentation, requires tags to figure out version

castelao:
When checking repository, include tags history to allow figuring out the current package version.

Check the version shown at https://nrel.github.io/sup3r/ to understand the issue.


bnb32 (approved):
Good catch!

castelao:
> Good catch!

Thanks!

PR #207, opened 2024-04-08
combined conditional moments and wind conditional moments.

bnb32:


grantbuster (approved):
Woah we didnt need any of those wind-specific moment batch classes? Is this because of the more general exo data handling you implemented previously? Awesome, love to see that -406 lines count! Less code == easier life.

bnb32:
> Woah we didnt need any of those wind-specific moment batch classes? Is this because of the more general exo data handling you implemented previously? Awesome, love to see that -406 lines count! Less code == easier life.

Yeah exactly!

PR #212, opened 2024-05-02
Basic setup for using pixi

castelao:
Setup to run on Eagle.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,6 @@ platforms = ["osx-arm64", "linux-64"]
 
 [tool.pixi.dependencies]
 python = "==3.11"
-netcdf4 = ">=1.5.8"
 cftime = ">=1.6.2"
 dask = ">=2022.0"
 h5netcdf = ">=1.1.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,3 +108,39 @@ packages = ["sup3r"]
 
 [tool.setuptools_scm]
 version_file = "sup3r/_version.py"
+
+[tool.pixi.project]
+channels = ["conda-forge", "anaconda", "main"]
+platforms = ["osx-arm64", "linux-64"]
+
+[tool.pixi.dependencies]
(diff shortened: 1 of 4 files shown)


castelao:
It's not worth to setup for Eagle considering it will be out soon. This setup shouldn't affect anyone but it would help me.

grantbuster (approved):
I'm still not familiar with the pixi setup so lgtm, but how are you using this? At this point i think i'm the only person running sup3r on eagle which is fine. You and brandon should focus on getting a kestrel gpu env setup with tensorflow version pinned. I'll finish model validation on Eagle and then we'll migrate things over when kestrel GPUs are available.

pyproject.toml
```diff
@@ -108,3 +108,39 @@ packages = ["sup3r"]
...
 [tool.setuptools_scm]
 version_file = "sup3r/_version.py"
+
+[tool.pixi.project]
+channels = ["conda-forge", "anaconda", "main"]
+platforms = ["osx-arm64", "linux-64"]
+
+[tool.pixi.dependencies]
+python = "==3.11"
+netcdf4 = ">=1.5.8"
```
bnb32, line 118: I dont think we need both netcdf and h5netcdf

bnb32 (changes requested):
looks good! small comment on netcdf vs h5netcdf. also, a key part of this is some additional instructions on use though. could you add some to the README?

castelao:
> ... also, a key part of this is some additional instructions on use though. could you add some to the README?

Yes. I'm planning to add the instructions in a following PR. With this one here I can do some tests in kestrel, and tune it, before recommending you both on how to use it.

bnb32:
> > ... also, a key part of this is some additional instructions on use though. could you add some to the README?
> 
> Yes. I'm planning to add the instructions in a following PR. With this one here I can do some tests in kestrel, and tune it, before recommending you both on how to use it.

Ok, that works.

PR #213, opened 2024-05-02
Fix badges on main README

ppinchuk:
Just a quick README update to align badges and not have them on separate lines. See new (rather - old) style here: https://github.com/NREL/sup3r/tree/pp/badges


bnb32 (approved):
Nice! Thanks for the fix

PR #216, opened 2024-05-20
Reorganizing QDM related methods

castelao:
The `bias_calc` was already getting too large, so let's split and reorganize it.

This PR is meant to simplify the review of the following `PresRat`. Here, it is mostly about moving things around. Any error would be due to paths accessing components of the library, but the actual existing methods are not changed. The `PresRat` will be implemented in `bias_calc.qdm`, on the side of the standard `QDM`.


castelao:
@grantbuster and @bnb32 , may I merge this one?

grantbuster (approved):

PR #217, opened 2024-06-05
fix performance issue with slow listcomp on regrid indexing

grantbuster:
@bnb32 this list comprehension was absolutely nuking the regridder performance for large timeseries


bnb32 (approved):
Yeah a similar fix is in my branch. Not sure why I didn't vectorize this initially.

PR #218, opened 2024-06-17
Migrating to kestrel using pixi to manage environment

castelao:
Kestrel setup to run sup3r using pixi to manage environment. The goal here is, in order of priority:
- Be able to run on kestrel.
- Keep control of the environment to guarantee reproducibility.
- Update tensorflow to 2.15. We are not compatible yet with 2.16.
- Sync dependency tree between linux and OSX for development

Some relevant notes:
- There is some issue with GLIBC, and one or more dependencies might have not been properly dynamically linked, so some combinations of versions create issues.
- A few tests are failing on kestrel and it's not clear the reasons. I'm addressing one case in another PR, but there are more.
- Using tensorflow 2.15.1, which is only available through pip.
- I couldn't make yet tensor flow-metal to work while keeping the dependency tree compatible. Which is probably OK for now since OSX would be for development only.


castelao:
A comment. This is not ideal yet, as it is, but might be the best choice at the moment to use kestrel. The main limitation now are the tests that fail. I wouldn't merge this until we understand what are those errors.

PR #220, opened 2024-06-23
Bnb/dh refactor

bnb32:
Ok, here we go...

`sup3r/preprocessing` was previously just data handlers and batch handlers, essentially. Now we have `Loaders`, `Extracters`, `Derivers`, `Cachers` which are composed in `sup3r.preprocessing.data_handlers.factory` to build objects similar to the old `DataHandlers`. These do basically everything the old handlers used to do, except for training / batching related routines like sampling, normalization, etc. `Loaders` just load netcdf / h5 data into a `xr.Dataset` - like container. `Extracters` extract spatiotemporal regions of data. `Derivers` derive new features from raw feature data. `Cachers`, well, they cache data to either h5 or netcdf depending on the extension of the output file provided. 

In `sup3r/preprocessing` we additionally have `Samplers` and `BatchQueues`. These are composed in `sup3r.preprocessing.batch_handlers.factory` to build objects similar to the old `BatchHandlers`. These do basically everything that the old batch handlers used to do, with some exceptions. The most notable exception is probably that they don't split data into training and validation sets. `BatchHandler` objects will take "collections" of data handler like objects (these can be `DataHandlers`, `Extracters`, `Derivers`, etc) for both training and validation and separate batch queues will be used for each. `Samplers` simply contain a `xr.Dataset` - like object and sample that data as an iterator. `BatchQueue` objects interface with samplers to keep a queue full of batches / samples while models are training.  

All these smaller objects like `loaders`, `extracters`, `derivers`, `samplers` are built on top of xr.Dataset - like objects (`sup3r.preprocessing.accessor.Sup3rX` and `sup3r.preprocessing.base.Sup3rDataset`) which serve as the familiar `.data` attribute for data and batch handlers. `Sup3rDataset` is wrapped around `Sup3rX` to provide an interface for "dual" dataset objects contained by dual handlers and acts exactly like `Sup3rX` when datasets are not dual. `Sup3rX` is an `xr.Dataset` "accessor" class, which is the recommended way to extend `xr.Datasets` (as opposed to subclassing). These `Sup3rX` / `Sup3rDataset` objects act similar to `xr.Datasets` but with extended functionality. The tests in `tests/data_wrappers/` show how to interact with these objects. 

Since the fundamental dataset objects are now `xr.Dataset` - like, they can use dask arrays to store data. This means we don't need to load data into memory until we need the result of a computation. `ForwardPassStrategy` and `ForwardPass` have been updated accordingly, since we can lazy load the full input dataset and then index the data handler `.data` attribute to select generator input chunks, all before loading into memory. `BatchHandler` objects have a `mode` argument which can be set to either `lazy` (load batches into memory only when they are sent out for training) or `eager` (load `.data` into memory upon handler initialization). 

Tests have been added for all new preprocessing modules and lots of documentation / notes have been added throughout. Tests should hopefully provide good examples of use patterns for these new objects.


sup3r/preprocessing/samplers/cc.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+        return high_res
+
+    @staticmethod
+    def get_middle(high_res, sample_shape):
+        """Get middle chunk of high_res data that will then be reduced to day
+        time steps. This has n_time_steps = 24 if sample_shape[-1] <= 24
+        otherwise n_time_steps = sample_shape[-1]."""
+        n_days = int(high_res.shape[3] / 24)
+        if n_days > 1:
```
grantbuster, line 133: What if `n_days == 1`? at least raise an explicit error maybe?

sup3r/preprocessing/samplers/cc.py
```diff
@@ -0,0 +1,179 @@
...
+            mid = int(np.ceil(high_res.shape[3] / 2))
+            start = mid - np.max((sample_shape[-1] // 2, 12))
```
grantbuster, line 135: This all makes sense but we may want to revisit how the nighttime hours get padded. Right now the hourly observations during nighttime are garbage. Would be better to fix them at the nearest valid values. Does that make sense? If you plot up the hr values from this test you'll see what i mean (screenshot below). I cant remember what we did for the previous solarcc model but i dont think this will work well. 

https://github.com/NREL/sup3r/pull/220/files#diff-5c2bda4e2d2e1efcea8d269d4b3f953453e8ca2a59876b60c6e15ee9fbef14a3R46

Timestep around sunrise from the test: 
image

sup3r/preprocessing/samplers/cc.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        if self.t_enhance not in (24, 1):
+            high_res = self.get_middle(high_res, self.hr_sample_shape)
+            high_res = nsrdb_reduce_daily_data(
+                high_res, self.hr_sample_shape[-1], csr_ind=csr_ind
+            )
+        return high_res
+
+    @staticmethod
+    def get_middle(high_res, sample_shape):
```
grantbuster, line 128: Call this `get_middle_days`

sup3r/preprocessing/samplers/cc.py
```diff
@@ -0,0 +1,179 @@
...
+        there is no need for reduction since every daily time step will have 24
+        hourly time steps in the high_res batch data. Of course, if t_enhance =
+        1, we are running for a spatial only model so this routine is
+        unnecessary.
+
+        *Needs review from @grantbuster
+        """
+        if self.t_enhance not in (24, 1):
+            high_res = self.get_middle(high_res, self.hr_sample_shape)
+            high_res = nsrdb_reduce_daily_data(
```
grantbuster, line 122: minor but in these methods (two places) why did you move np.round -> np.ceil?

grantbuster (changes requested):

sup3r/preprocessing/samplers/dual.py
```diff
@@ -0,0 +1,139 @@
...
+        s_enhance: int = 1,
+        t_enhance: int = 1,
+        feature_sets: Optional[Dict] = None,
+    ):
+        """
+        Parameters
+        ----------
+        data : Sup3rDataset
+            A tuple of xr.Dataset instances. The first must be low-res
+            and the second must be high-res data
```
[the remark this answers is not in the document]
bnb32: Sup3rDataset is basically a wrapped named tuple but I'll make this more clear.

sup3r/utilities/pytest/helpers.py
```diff
@@ -0,0 +1,405 @@
...
+    """Build test samplers which track indices."""
+
+    class SamplerTester(SamplerClass):
+        """Keep a record of sample indices for testing."""
+
+        def __init__(self, *args, **kwargs):
+            self.index_record = []
+            super().__init__(*args, **kwargs)
+
+        def get_sample_index(self, **kwargs):
```
grantbuster, line 130: Can we consider just adding this into the default batch handler infrastructure? Can be nice to know the historical sample indices and will simplify things a bit. Shouldn't be too much compute? We could always just store the index record for only the most recent sample.

sup3r/preprocessing/samplers/cc.py
bnb32: Yeah I thought about just limiting sample_shape[2] <= 24 for this. It seems that the previous behavior essentially defaulted to this, since the max size of the time dimension returned by reduce_high_res_sub_daily used to be 24. Would that accomplish what we want here?

sup3r/preprocessing/samplers/cc.py
bnb32: Yeah I think it shouldnt make it here if the enhancement factor results in high_res.shape[3] == 24. I think that would require t_enhance = 24 and sample_shape[2] = 24. Could do a check on the given t_enhance value to account for this

sup3r/preprocessing/samplers/cc.py
bnb32: So I could account for an odd index for a mid point

sup3r/preprocessing/samplers/base.py
```diff
@@ -0,0 +1,316 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Sampler(Container):
+    """Sampler class for iterating through samples of contained data."""
+
+    def __init__(
+        self,
+        data: T_Dataset,
```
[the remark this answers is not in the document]
bnb32: Yeah thats fair. I started trying to be a little fancy with types but this one prob isnt helpful.

sup3r/utilities/pytest/helpers.py
bnb32: I didn't want to do this bc itll likely go unused and mostly only helps with testing. We added some stuff last time to make testing easier and it resulted in clutter.

sup3r/preprocessing/samplers/cc.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        See Also
+        --------
+        :class:`DualSampler` for argument descriptions.
+        """
+        msg = (
+            f'{self.__class__.__name__} requires a Sup3rDataset object '
+            'with `.daily` and `.hourly` data members, in that order'
+        )
+        assert hasattr(data, 'daily') and hasattr(data, 'hourly'), msg
```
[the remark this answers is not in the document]
bnb32: Yeah will do. You just initialize like Sup3rDataset(daily=xr.Dataset(), hourly=xr.Dataset()). These kwargs are used as the names in a namedtuple so they can be anything (e.g. low_res, high_res).

sup3r/preprocessing/base.py
```diff
@@ -0,0 +1,414 @@
...
+    temperature_2m, temperature_max_2m, temperature_min_2m). This means all
+    these variables should have the same means and stds, which ultimately come
+    from the high_res non coarsened variable.
+    """
+
+    def __init__(
+        self,
+        data: Optional[Union[tuple, T_Dataset]] = None,
+        **dsets: Union[xr.Dataset, Sup3rX],
+    ):
```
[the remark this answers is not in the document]
bnb32: Yeah Ill add some stuff here. Im confused on the naming recs though? Would data_hr be a singlet and data_hr_lr a doublet?

sup3r/postprocessing/writers/nc.py
```diff
@@ -0,0 +1,132 @@
...
+            Dictionary of meta data from model
+        max_workers: None | int
+            Has no effect. Compliance with parent signature.
+        gids : list
+            List of coordinate indices used to label each lat lon pair and to
+            help with spatial chunk data collection
+        """
+        coords = {
+            Dimension.TIME: times,
+            Dimension.SOUTH_NORTH: lat_lon[:, 0, 0],
```
castelao, line 60: A question. It is always guarantee that latitude repeat along the first dimension and longitude repeat along the second dimension, i.e. `lat_lon[:,0,0] == lat_lon[:,1,0]`

castelao (commented):
Just a few comments. I didn't have the chance to check much of your PR, but it looks great so far.

pyproject.toml
```diff
@@ -294,10 +296,19 @@ sphinx_rtd_theme = ">=2.0"
...
 [tool.pixi.feature.dev.dependencies]
 build = ">=0.6"
 twine = ">=5.0"
 ruff = ">=0.4"
 ipython = ">=8.0"
 pytest-xdist = ">=3.0"
+
+[tool.pixi.feature.viz.dependencies]
+jupyter = ">1.0.0"
+hvplot = ">0.10.0"
```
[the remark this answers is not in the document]
bnb32: cool, thanks for catching these

sup3r/models/solar_cc.py
```diff
@@ -1,18 +1,19 @@
...
 
-
 logger = logging.getLogger(__name__)
 
 
 class SolarCC(Sup3rGan):
     """Solar climate change model.
 
-    Modifications to standard Sup3rGan:
+    Note
```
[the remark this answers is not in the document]
bnb32: Oh, do we have to change something for sphinx to render correctly? I changed it because we werent getting a nice blue box around Notes

sup3r/postprocessing/writers/nc.py
bnb32: No and this is definitely not the case for WTK or NSRDB data, thats why I keep the 2d array for lat / lon coords. I actually just added south_north / west_east for easier plotting with hvplot, but then I found `.quadmesh` so I can remove this.

sup3r/bias/mixins.py
```diff
@@ -11,12 +11,14 @@
...
 class FillAndSmoothMixin:
-    """Fill and extend parameters for calibration on missing positions"""
-    def fill_and_smooth(self,
-                        out,
-                        fill_extend=True,
-                        smooth_extend=0,
-                        smooth_interior=0):
+    """Fill and extend parameters for calibration on missing positions
+
+    TODO: replace nn_fill_array call with `Sup3rX.interpolate_na` method
```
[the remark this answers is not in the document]
bnb32: Yeah good suggestion. These TODOs were mostly for me so I wouldnt forget but I leave this alone for now and add an issue.

sup3r/preprocessing/accessor.py
```diff
@@ -0,0 +1,527 @@
...
+    ordered_dims,
+    parse_to_list,
+)
+from sup3r.typing import T_Array
+
+logger = logging.getLogger(__name__)
+
+
+@xr.register_dataset_accessor('sx')
+class Sup3rX:
```
grantbuster, line 32: can you add a `Sup3rX.values` property that is basically `np.asarray(self[...])`? Not sure if .values would be useful elsewhere but you should consider it... really nice to be able to extract the numpy array from xarray arrays.

grantbuster (changes requested):

sup3r/preprocessing/accessor.py
bnb32: You can already do `Sup3rX[dset].values` since it acts like an `xr.Dataset` with extras.

sup3r/preprocessing/loaders/base.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        Parameters
+        ----------
+        file_paths : str | pathlib.Path | list
+            Location(s) of files to load
+        features : list | str
+            Features to return in loaded dataset. If 'all' then all available
+            features will be returned.
+        res_kwargs : dict
+            kwargs for `.res` object
```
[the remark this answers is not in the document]
bnb32: Done, thanks

sup3r/models/solar_cc.py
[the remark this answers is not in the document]
castelao: Oh, we might be talking about different things. The [Notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) is one of the sections, such as `Returns` or `Examples`.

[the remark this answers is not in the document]
bnb32: Oh yeah, different things.

sup3r/preprocessing/base.py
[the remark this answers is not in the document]
grantbuster: Yeah you basically have handling for multiple options for data, right? if data is tuple do this... if data is a single do this... I'm saying you should explicitly provide kwargs data_hr which would be a single interpreted as the high-res data or data_hr_lr/data_pair which would be the high-res/low-res tuple. This makes it more explicit what you expect and what the priorities are... I think we should prefer explicit kwarg options rather than describing the options in the docstring for a single arg. 

Separate note, in this docstring you say dsets is the prefered option as a dictionary... You should say it's flexible kwargs and then name the keys. The example in the class doc really explains this better than the dsets docstring. 

Last note: i don't think the word "singleton" should be used in this case. This has a specific definition in design patterns that doesn't match what you're describing.

sup3r/preprocessing/samplers/base.py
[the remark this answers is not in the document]
grantbuster: you still have T_Array as a dtype in some places - i think same comment. define this somewhere easily accessible or just write out the union statement

sup3r/preprocessing/samplers/cc.py
grantbuster: should probably be stopped somewhere, either in the init or in this function. Can you just add an assert statement?

sup3r/preprocessing/samplers/cc.py
grantbuster: I actually think we're all good here. The sampling is working as designed and the reason for the weird patterns in the first image is just that the entire scene has GHI < 10 which leads to weird looking clearsky ratios. I added an issue with suggestions for ways to improve this but we don't need to do anything for this right now

https://github.com/NREL/sup3r/issues/229

sup3r/preprocessing/samplers/cc.py
grantbuster: wouldn't round do this? This is maybe a moot point because the high res shape should always be a multiple of 24.

sup3r/utilities/pytest/helpers.py
grantbuster: Just removing the testing code. I like having some way to pull the observation indices when verifying my training setup. I can use the test handlers but seems like a lightweight feature that could be easily included in the base handlers. I don't feel strongly about this so no worries either way, it's just a nice feature.

sup3r/preprocessing/accessor.py
grantbuster: yeah but `np.asarray(self[...])` does a nice job of stacking the feature channel... can't we just add the .values property? pleaseeeee

sup3r/preprocessing/base.py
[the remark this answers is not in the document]
bnb32: Yeah good call on the terminology. The reason I didnt make it explicit is because the `.data` member of all container classes gets wrapped as a `Sup3rDataset` through `Container@setter.data`. This includes containers with a single data member and dual containers. I'd have to put these checks in multiple places if they were removed from this init.

sup3r/preprocessing/samplers/cc.py
bnb32: Yeah maybe round would do this also. High res doesn't need to be a multiple of 24 though. see https://github.com/NREL/sup3r/blob/a6acb905c310c75908dd899e8725203791e0c115/tests/batch_handlers/test_bh_h5_cc.py#L31. I guess before it goes through reduce daily it is always a multiple of 24 though.

sup3r/preprocessing/samplers/base.py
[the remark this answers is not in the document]
bnb32: Yeah I discovered the reason I did this for T_Dataset. Past Brandon had thought through it apparently :). When I import Sup3rX and Sup3rDataset everywhere I added a T_Dataset type hint it caused circular imports. This one is easier to replace though.

sup3r/preprocessing/samplers/cc.py
bnb32: This can actually happen also if sample_shape[2] = 24 and t_enhance = 1. Which should be allowed right? Im confused why n_days < 1 and n_days > 1 but not n_days = 1 should be allowed here

sup3r/preprocessing/accessor.py
bnb32: I added this in the latest update

sup3r/utilities/pytest/helpers.py
bnb32: For now I think Ill leave it. If you find that you are constantly having to use the test handlers lets add it in.

sup3r/preprocessing/samplers/base.py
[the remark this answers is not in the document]
bnb32: all custom types removed now

grantbuster (approved):

PR #222, opened 2024-06-25
fixed bias calc class import path

grantbuster:


castelao:
Good catch! I missed that re-routing when I created the `.qdm`.

PR #215, opened 2024-05-06
PresRat correction

castelao:
Implementing PresRat for precipitation.

PR #224, opened 2024-06-27
Improving ruff setup

castelao:
Isolating ruff setup into its own PR to make it easier to review. PR #215 missed a few criteria without warning and hopefully we'll be able to identify those after this PR.

These are not my choices. I reviewed all rules to best match the former Pylint setup. We don't have a perfect match on the rules from Pylint vs Ruff, so it is the closest as possible. After that, a few standard rules were added, and can be easily removed. Several specific rules were included to conform with the current code, and some of those might be worth re-consider and conform.

By using `pyproject.toml` to setup, it should unify the checks, i.e. the checks on GA should be identical to the local run, which was not true with the previous pylint setup.

This new check was added with the GA workflow together with the former procedure. We might keep that one for some time for comparison. We most probably will find different checking results which will guide the choices for this repository. Eventually, pylint and flake8 could be removed.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,3 +28,13 @@ jobs:
           VALIDATE_YAML: false
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  ruff:
+    name: Ruff
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: chartboost/ruff-action@v1
+        with:
+          version: 0.4.10
+          args: check
+          src: "./sup3r ./tests"
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches-ignore:
       - 'gh-pages'
+  pull_request:
+    branches: [main]
 
 jobs:
   build:
@@ -17,7 +19,7 @@ jobs:
       - name: Lint Code Base
         uses: super-linter/super-linter@v4
         env:
-          VALIDATE_ALL_CODEBASE: false
+          VALIDATE_ALL_CODEBASE: true
           VALIDATE_PYTHON_BLACK: false
           VALIDATE_PYTHON_ISORT: false
           VALIDATE_PYTHON_MYPY: false
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Lint Code Base
         uses: super-linter/super-linter@v4
         env:
-          VALIDATE_ALL_CODEBASE: true
+          VALIDATE_ALL_CODEBASE: false
           VALIDATE_PYTHON_BLACK: false
           VALIDATE_PYTHON_ISORT: false
           VALIDATE_PYTHON_MYPY: false
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -36,5 +36,5 @@ jobs:
       - uses: chartboost/ruff-action@v1
         with:
           version: 0.4.10
-          args: check
+          args: check --output-format=github
           src: "./sup3r ./tests"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,7 +99,7 @@ select = [
   "I", # isort
   "NPY", # numpy-specific
   "Q", # flake8-quotes
-#  "UP", # pyupgrade
+  "UP", # pyupgrade
   "W", # Warning
   ]
 
@@ -137,8 +137,9 @@ ignore = [
   "PLW1514", # unspecified-encoding
   "Q000", # bad-quotes-inline-string
   "Q004", # unnecessary-escaped-quote
-#  "UP009", # utf8-encoding-declaration
-#  "UP032", # f-string
+  "UP009", # utf8-encoding-declaration
+  "UP015", # redundant-open-modes
+  "UP032", # f-string
 #  "UP038" # non-pep604-isinstance
   ]
 # Ignored in pylint setup but missing on ruff. We shall delete from the
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,7 @@ select = [
   "NPY", # numpy-specific
   "PL", # Pylint
   "Q", # flake8-quotes
+  "SIM", # flake8-simplify
   "UP", # pyupgrade
   "W", # Warning
   ]
@@ -159,6 +160,10 @@ ignore = [
   "PLW2901", # redefined-loop-name
   "Q000", # bad-quotes-inline-string
   "Q004", # unnecessary-escaped-quote
+  "SIM108", # if-else-block-instead-of-if-exp
+  "SIM117", #multiple-with-statements
+  "SIM118", # in-dict-keys
+  "SIM211", # if-expr-with-false-true
   "UP009", # utf8-encoding-declaration
   "UP015", # redundant-open-modes
   "UP032", # f-string
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ select = [
 #  "NPY", # numpy-specific
 #  "Q", # flake8-quotes
 #  "UP", # pyupgrade
-#  "W", # Warning
+  "W", # Warning
   ]
 
 ignore = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,10 +90,7 @@ select = [
   "COM", # flake8-commas
   "D", # pydocstyle
   "E", # pycodestyle
-  "E701", # multiple-statements-on-one-line-colon
-  "E722", # bare-except
   "F", # Pyflakes
-  "F401", # unused-import
   "G", # flake8-logging-format
   "I", # isort
   "LOG", # flake8-logging
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ select = [
   "A", # flake8-builtins
   "ARG", # flake8-unused-arguments
   "COM", # flake8-commas
-#  # "D", # pydocstyle
+  "D", # pydocstyle
   "E", # pycodestyle
   "E701", # multiple-statements-on-one-line-colon
   "E722", # bare-except
@@ -114,14 +114,18 @@ ignore = [
 #  "B028", # no-explicit-stacklevel
 #  "B905", # zip-without-explicit-strict
   "COM812", # missing-trailing-comma
-#  "D105", # undocumented-magic-method
-#  "D202", # no-blank-line-after-function
-#  "D205", # blank-line-after-summary
-#  "D209", # new-line-after-last-paragraph
+  "D105", # undocumented-magic-method
+  "D200", # fits-on-one-line
+  "D202", # no-blank-line-after-function
+  "D204", # one-blank-line-after-class
+  "D205", # blank-line-after-summary
+  "D207", # under-indentation
+  "D209", # new-line-after-last-paragraph
 #  "D212", # multi-line-summary-first-line
 #  "D213", # multi-line-summary-second-linek
-#  "D400", # ends-in-period
-#  "D401", # non-imperative-mood
+  "D400", # ends-in-period
+  "D401", # non-imperative-mood
+  "D404", # docstring-starts-with-this
 #  "D413", # blank-line-after-last-section
 #  "D415", # ends-in-punctuation
 #  "E902", # io-error
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ exclude = [
 [tool.ruff.lint]
 # fixable = ["ALL"]
 # preview = true
+task-tags = ["TODO", "FIXME", "XXX"]
 select = [
 #  "E", # pycodestyle
 #  "F", # Pyflakes
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,6 +189,7 @@ ignore = [
 #  ]
 
 [tool.ruff.lint.pylint]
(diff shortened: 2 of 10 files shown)


castelao:
Note the time difference of 9s vs ~90s, but the checks don't match so it's not a fair comparison.

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
 ignore = [
-  "B008", # function-call-in-default-argument
-  "B024", # abstract-base-class-without-abstract-method
-  "B028", # no-explicit-stacklevel
-  "B905", # zip-without-explicit-strict
+  # Currently don't conform but we might want to reconsider
+  "A001", # builtin-variable-shadowing
+  # Currently don't conform but we might want to reconsider
+  "A002", # builtin-argument-shadowing
+  "ARG002", # unused-method-argument
```
bnb32, line 109: I think we should keep all but ARG005 as default, and require line specific overrides. I think this is basically what weve been doing for pylint

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
   "D105", # undocumented-magic-method
+  "D200", # fits-on-one-line
   "D202", # no-blank-line-after-function
+  "D204", # one-blank-line-after-class
   "D205", # blank-line-after-summary
+  "D207", # under-indentation
   "D209", # new-line-after-last-paragraph
-  "D212", # multi-line-summary-first-line
-  "D213", # multi-line-summary-second-linek
+#  "D212", # multi-line-summary-first-line
```
bnb32, line 125: Why did you comment these ones out?

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+  "E722", # bare-except
   "F", # Pyflakes
+  "F401", # unused-import
+  "G", # flake8-logging-format
+  "I", # isort
+  "N", # pep8-naming
+  "NPY", # numpy-specific
+  "PL", # Pylint
+  "Q", # flake8-quotes
+  "UP", # pyupgrade
```
bnb32, line 100: I love using UP, and will continue to locally, but maybe not needed for the standard rule set? I could go either way on that. What do you think?

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
-fixable = ["ALL"]
-preview = true
+fixable = []
+# preview = true
+# logger-objects = []
+task-tags = ["TODO", "FIXME", "XXX"]
 select = [
+  "A", # flake8-builtins
+  "ARG", # flake8-unused-arguments
+  "COM", # flake8-commas
```
bnb32, line 87: Does "A" not cover what we want to keep from flake? I dont use COM in my local rule set but maybe I should be?

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+  "D", # pydocstyle
   "E", # pycodestyle
+  "E701", # multiple-statements-on-one-line-colon
+  "E722", # bare-except
   "F", # Pyflakes
+  "F401", # unused-import
+  "G", # flake8-logging-format
+  "I", # isort
```
bnb32, line 95: isort just has two rules so we could just include I002 here and remove I001 from ignores.

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+  "N", # pep8-naming
+  "NPY", # numpy-specific
+  "PL", # Pylint
+  "Q", # flake8-quotes
+  "UP", # pyupgrade
   "W", # Warning
   ]
```
bnb32, line 102: I think we should include C901 here also (too-complex)

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+]
 
 [tool.ruff.lint]
-fixable = ["ALL"]
-preview = true
+fixable = []
+# preview = true
+# logger-objects = []
+task-tags = ["TODO", "FIXME", "XXX"]
 select = [
```
bnb32, line 84: Whatare your thoughts on including / not-including "C4" (flake8-comprehensions")

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+  "PLR1702", # too-many-nested-blocks
+  "PLR1704", # redefined-argument-from-local
+  "PLR2004", # magic-value-comparison
+  "PLW1514", # unspecified-encoding
+  "PLW2901", # redefined-loop-name
+  "Q000", # bad-quotes-inline-string
+  "Q004", # unnecessary-escaped-quote
   "UP009", # utf8-encoding-declaration
+  "UP015", # redundant-open-modes
   "UP032", # f-string
```
bnb32, line 158: this one is for @grantbuster ;)

pyproject.toml
```diff
@@ -71,49 +71,161 @@ line-length = 79
...
+max-locals=15 # (PLR0912) local variables allowed for a function or method body
+max-nested-blocks = 5 # (PLR1702) nested blocks within a function or method body
+max-public-methods=20 # (R0904) public methods allowed for a class
+max-returns=6 # (PLR0911) return statements for a function or method body
+max-statements=50 # (PLR0915) statements allowed for a function or method body
 
 [tool.ruff.format]
 quote-style = "single"
 indent-style = "space"
+# Consider adopting "lf" instead
```
bnb32, line 227: whats "lf"?

bnb32 (changes requested):
Love that were doing this and thanks for getting this started so quickly. A few questions / comments

pyproject.toml
castelao: I think it is a good idea, but as it is, if I don't ignore those, it gives errors.

pyproject.toml
castelao: I've also been using ruff from the beginning, and slowly changing the rules to conform with the sup3r standard as I get errors. At some point `D212` was a requirement. Yesterday I removed everything and checked each one. As it is, sup3r pass without ignoring `D212`. I left commented while we double check everything with the intention to remove soon.

pyproject.toml
castelao: I suggest using it. The first time I used was a big surprise for myself. I've been using Python for so long that I didn't realize how I was getting outdated. That should help us to be aware of how the Python syntax is evolving. It will make a difference once we bump the minimum Python version.

That said, whatever you both like, works for me.

pyproject.toml
castelao: Yes we can. I suggest using this way. If any new rule is implemented in ruff's 'I' section, we get it; This reminds us that we ignore `I001` and might want to consider one day to conform with it.

pyproject.toml
castelao: I'll add it.

pyproject.toml
bnb32: Yeah well have to add inline ruff overrides along side the pylint ones

pyproject.toml
bnb32: Yeah thats a good approach

pyproject.toml
bnb32: As per our discussion, well remove these ignores in the next PR

pyproject.toml
bnb32: Agreed. Lets keep it

pyproject.toml
castelao: Windows has a different end of line standard with an extra special character, using '\r\n\' (cr-lf pattern) while nix OS uses '\n' (lf pattern). Conforming to one standard helps to avoid strange characters showing in OSX & Linux.

BTW, on windows machine it is probably a good idea to setup global just once: `git config --global core.autocrlf true`, if you haven't already done that.

bnb32: Oh, good to know! I've had many annoying experiences with unix vs windows path standards

pyproject.toml
castelao: Done!

grantbuster:
hey @bnb32 and @castelao, nice work and thanks for prioritizing this. There's too many codes and new configs for me to learn and review right now but seems like you two have been having a productive dialogue which is what matters. Let's get this merged in when you two are happy and see what it does with the open PRs and we can tweak as we go.

pyproject.toml
castelao: Done!

bnb32:
@castelao One more thing. Lets add `on: workflow_dispatch` to the actions so we can manually trigger this when we want to.

.github/workflows/linter.yml
```diff
@@ -28,3 +28,13 @@ jobs:
           VALIDATE_YAML: false
```
grantbuster, line 28: Actually one comment, are we keeping superlinter + ruff? Maybe remove super linter? Or is it so broken right now that it's not even running?

bnb32: We were thinking that it should be kept until the presrat merge (and gui would figure out the pylint issue) so we didnt break too many things. Some inline overrides will need to be added to appease ruff when we fully adopt

pyproject.toml
castelao: We don't break it, why not using COM?

I added a couple more to better cover flake8. Anything specific you would like to include?

castelao:
> @castelao One more thing. Lets add `on: workflow_dispatch` to the actions so we can manually trigger this when we want to.

Done!

pyproject.toml
bnb32: Was just curious if its redundant. I'm fine with including it though. Re: other includes - I have SIM and PERF in my own set but this will break too much. Otherwise this closely matches my config so I'm happy.

.github/workflows/linter.yml
castelao: Good point. We don't have a match 1:1 between both tools. I was planning on keeping superlinter for a short time to see what they disagree so we can create an opinion on what is missing and should be included or dropped.

I don't know what is the issue with superlinter right now, but it don't hurt to keep it for a short time. I'll move to PresRat after this, and can return to superlinter after that.

castelao:
> hey @bnb32 and @castelao, nice work and thanks for prioritizing this. There's too many codes and new configs for me to learn and review right now but seems like you two have been having a productive dialogue which is what matters. Let's get this merged in when you two are happy and see what it does with the open PRs and we can tweak as we go.

Without going in details on the rules, I tried to reproduce the best I could the previous setup, but there is not always a 1:1 match. After that, I added some standard rules that don't break the current sup3r, which means that those rules were already followed. In summary, there is no change in the pattern of sup3r here, but instead we should have rules that match sup3r.

Some specific rules I had to ignore, and a few of them I added a suggestion to review the decision.

grantbuster:
> > hey @bnb32 and @castelao, nice work and thanks for prioritizing this. There's too many codes and new configs for me to learn and review right now but seems like you two have been having a productive dialogue which is what matters. Let's get this merged in when you two are happy and see what it does with the open PRs and we can tweak as we go.
> 
> Without going in details on the rules, I tried to reproduce the best I could the previous setup, but there is not always a 1:1 match. After that, I added some standard rules that don't break the current sup3r, which means that those rules were already followed. In summary, there is no change in the pattern of sup3r here, but instead we should have rules that match sup3r.
> 
> Some specific rules I had to ignore, and a few of them I added a suggestion to review the decision.

Sounds great 👍

pyproject.toml
castelao: > Windows has a different end of line standard with an extra special character, using '\r\n' (cr-lf pattern) while nix OS uses '\n' (lf pattern). Conforming to one standard helps to avoid strange characters showing in OSX & Linux.
> 
> BTW, on windows machine it is probably a good idea to setup global just once: `git config --global core.autocrlf true`, if you haven't already done that.

Sorry, Just to be clear. That command is for windows only.

pyproject.toml
castelao: I'm closing this 'conversation'. Please, feel free to re-open if we're missing anything.

pyproject.toml
castelao: @bnb32 , I just added those two. I had to ignore a few specific rules so it doesn't break with sup3r as it is. As I mentioned before, I suggest starting like that and later we can conform and remove the specific violations one by one, but it would work as it is now.

castelao: I'm closing this one but feel free to re-open if want to continue this discussion.

pyproject.toml
castelao: I'll close this one as well.

castelao:
@bnb32 @grantbuster , new rules setup is working fine. This is just a start. We can later review each one of the specific rules been ignored, but it runs as it is now.

There is just one strange error that the former super-linter is complaining in a yaml file that I didn't edit in this PR. It's not clear to me what is the issue there. I suggest we ignore that and move on to merge this PR. I think this is a great start and I really wanted to get back to PresRat.

.github/workflows/linter.yml
castelao: It running now, and the only complain is with a yaml file and I couldn't understand what is the issue.

castelao: Fixed! Closing this one.

castelao:
@grantbuster @bnb32 , passing all tests. Just waiting on your review.

bnb32:
> @bnb32 @grantbuster , new rules setup is working fine. This is just a start. We can later review each one of the specific rules been ignored, but it runs as it is now.
> 
> There is just one strange error that the former super-linter is complaining in a yaml file that I didn't edit in this PR. It's not clear to me what is the issue there. I suggest we ignore that and move on to merge this PR. I think this is a great start and I really wanted to get back to PresRat.

Sounds good to me. What yaml file is it complaining about? yaml files should be excluded from linting anyway imo

bnb32 (approved):

castelao:
@grantbuster , I guess you're also OK with this PR? Since @bnb32 already approved, I'll merge this one so I can use it in PresRat.

PR #225, opened 2024-07-02
Explicit keyword argument

castelao:
The forward pass uses `lat_lon` as a positional argument, which caused an error in the PresRat implementation. Is there a good reason to don't be explicit here by using a keyword argument?

I split this from #215 to be sure that this is not breaking anything else.


bnb32 (approved):
This seems fine to me. I prefer explicit kwargs anyway.

grantbuster (approved):
fine with me!

PR #228, opened 2024-07-29
Use `rex Regridder` port

ppinchuk:
`Regridder` was [ported to `rex`](https://github.com/NREL/rex/pull/176), so this PR deprecates the implementation in `sup3r` in favor of `rex`.

Thanks for sharing this functionality! :)


bnb32 (approved):
Code removed is good in my book! Thanks for moving this over.

PR #230, opened 2024-08-27
Gb/presrat updates 2

grantbuster:
1. Moved a few things around to make presrat callable from a script with parameters. 
2. Lean on `delta_demon_min` from rex to prevent divide by small numbers in QDM
3. window_size and n_time_steps are now kwargs so you can set at runtime. found issues with default class attrs for precip
4. fixed zero rate threshold to actually be 0.01 mm/day and updated use to prevent QDM divide by small number


tests/rasterizers/test_rasterizer_caching.py
```diff
@@ -57,11 +57,8 @@ def test_data_caching(input_files, ext, shape, target, features):
             rasterizer, cache_kwargs={'cache_pattern': cache_pattern}
         )
 
-        assert rasterizer.shape[:3] == (
-            shape[0],
-            shape[1],
-            rasterizer.shape[2],
-        )
+        good_shape = (shape[0], shape[1], rasterizer.shape[2])
```
bnb32, line 60: lol. fair point :)

sup3r/postprocessing/writers/base.py
```diff
@@ -105,6 +105,22 @@
         'min': -200,
         'max': 100,
     },
+    'temperature_min': {
```
bnb32, line 108: we should probably move this dict to its own file, huh?

sup3r/bias/qdm.py
```diff
@@ -69,9 +63,11 @@ def __init__(self,
                  match_zero_rate=False,
                  n_quantiles=101,
                  dist='empirical',
-                 relative=None,
+                 relative=True,
                  sampling='linear',
                  log_base=10,
+                 n_time_steps=24,
+                 window_size=120,
```
bnb32, line 70: I think this class needs a few more arguments ;)

bnb32 (approved):
I think this looks fine, esp for merging into the refactor branch, but we should be wary of letting the bc stuff get too bloated. Not sure if you get the same feeling that a clean up might be around the corner.

tests/rasterizers/test_rasterizer_caching.py
grantbuster: haha i dont even remember why i felt the need to fix this but yes trying to have less ugly one liners given the new linting.

sup3r/postprocessing/writers/base.py
grantbuster: Yes could be json just need to make sure the file gets included in pip install like this: https://github.com/NREL/reV/blob/main/setup.py#L87

Want me to do this or have you done it already?

sup3r/bias/qdm.py
grantbuster: Haha too much bloat? Possibly, but for now we just need to be able to modify these from config.

sup3r/postprocessing/writers/base.py
bnb32: Haven't done this yet so if you could that'd be awesome.

grantbuster: Okay done, can you merge when happy?

bnb32: will do. thanks!

PR #234, opened 2024-10-22
fixed double bias correction of features in bc qdm utility

grantbuster:


bnb32 (approved):
Oops. Missed the copy over from ``lin_bc``. Thanks!

PR #238, opened 2024-10-24
added simple test on cc batching for daily boundaries

grantbuster:


bnb32 (approved):
good addition. don't worry about the other failing tests right now.

PR #242, opened 2024-11-13
Gb/optm state

grantbuster:


bnb32 (approved):
Looks good. Any interesting results yet?

grantbuster:
> Looks good. Any interesting results yet?

not really, i think we'll need more training time to see if there's a trajectory. Some layers are definitely gaining momentum on gradients but not sure if they'll converge.

sup3r/preprocessing/cachers/base.py
```diff
@@ -475,7 +476,16 @@ def write_netcdf(
...
+                attr_value = safe_cast(attr_value)
+                try:
+                    ncfile.setncattr(attr_name, attr_value)
+                except Exception as e:
+                    msg = (f'Could not write {attr_name} as attribute, '
+                           f'serializing with json dumps, '
+                           f'received error: "{e}"')
+                    logger.warning(msg)
+                    warn(msg)
+                    ncfile.setncattr(attr_name, json.dumps(attr_value))
```
grantbuster, line 488: @bnb32  fyi i was getting a weird error in production runs where the model meta data had objects that were not compatible with NetCDF. I added some extra error handling here.

bnb32: ok cool. you should use the ``safe_serialize`` function (https://github.com/NREL/sup3r/blob/e43e4bbacfeceaf453e3fcff685a3323ba23745f/sup3r/utilities/utilities.py#L69) instead of json.dumps here.

grantbuster: smart, done!

PR #246, opened 2024-11-20
Gb/disc valid

grantbuster:
@bnb32 fyi i started this yesterday and already see good results with the solar model that has a small number of timesteps passed to the disc. This makes a ton of sense since stride=2 for a lot of the conv layers in disc, which will compress the image significantly and padding zeros will have an outsized effect on the small solar observation. The effect will become negligible as the observation size grows but i still think this is a worthwhile improvement.


bnb32 (approved):

bnb32:
> @bnb32 fyi i started this yesterday and already see good results with the solar model that has a small number of timesteps passed to the disc. This makes a ton of sense since stride=2 for a lot of the conv layers in disc, which will compress the image significantly and padding zeros will have an outsized effect on the small solar observation. The effect will become negligible as the observation size grows but i still think this is a worthwhile improvement.

Yeah this is a great point. The typical use case for the solar model (few timesteps) makes this definitely worth doing.

PR #247, opened 2024-11-25
Bnb/1d nc loading

bnb32:
Some adjustments to loaders so they work nicer with flattened nc data. Also fix for slicer edge case where `modulo(grid_shape, fwp_chunk_shape)` is too small for generator padding layer requirements.

### DIFF
--- a/sup3r/pipeline/slicer.py
+++ b/sup3r/pipeline/slicer.py
@@ -72,6 +72,7 @@ class ForwardPassSlicer:
         self.time_slice = _parse_time_slice(self.time_slice)
 
         self._chunk_lookup = None
+        self._extra_padding = None
         self._s1_lr_slices = None
         self._s2_lr_slices = None
         self._s1_lr_pad_slices = None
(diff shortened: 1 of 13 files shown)


sup3r/pipeline/slicer.py
```diff
@@ -494,11 +520,50 @@ def get_padded_slices(slices, shape, enhancement, padding, step=None):
...
+        # last slice adjustment
+        if 2 * self.spatial_pad + (lr_slice_stop - lr_slice_start) < 4:
+            logger.warning(
+                warn_msg,
+                dim + 1,
+                lr_slice_start,
+                lr_slice_stop,
+                self.spatial_pad,
+            )
+            warn(
```
grantbuster, line 559: I feel like this should almost be an exception and force you to adjust your fwp shape... It could result in some pretty garbage data. Does this increase the chunk size for the last chunk and crop more of the output? And leave the neighboring chunk alone? Actually that makes sense. Does this check vertical chunk dims too?

grantbuster (approved):

sup3r/pipeline/slicer.py
bnb32: Yeah this pads only the "final" chunk and then crops the output more. This is done for each spatial dimension so it ends up padding the right-most and bottom-most boundary. True edge case here since it almost never happens when you have non-zero spatial padding and then only happens if spatial padding is 1 and the last unpadded chunk size is also 1.

grantbuster: Gotcha that makes sense. good solution too!

PR #249, opened 2024-12-12
Bug fix on GCM clearsky ratio

grantbuster:
@bnb32, possibly a bit premature as i haven't tested thoroughly yet, but since kestrel is down and i know you're bored i thought i'd submit this for review now. 

Main fix is that `da.repeat` does not do what we think it does for multi year solar timeseries (see lines referenced below). We need to stack multiple years of the NSRDB clearsky ghi but da.repeat does [1, 2, 3] -> [1, 1, 2, 2, 3, 3] so you get one giant annual cycle which results in a lot of ratio=1 values at the edges. Leap days also matter... if we do nyears*366 for nyears>10 we end up with a growing shift in the later years. This matters a lot for the 40-year record we analyze for bias correction. 

https://github.com/NREL/sup3r/compare/main...gb/csratio#diff-d97151971c312b63001f37eafc4f5fac2020d7b0a7421d7b3e4adfa1bf7d1cebR223-R229

### DIFF
--- a/sup3r/preprocessing/data_handlers/nc_cc.py
+++ b/sup3r/preprocessing/data_handlers/nc_cc.py
@@ -3,7 +3,6 @@
 import logging
 import os
 
-import dask.array as da
 import numpy as np
 import pandas as pd
 from scipy.spatial import KDTree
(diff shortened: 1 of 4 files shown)


sup3r/preprocessing/data_handlers/nc_cc.py
```diff
@@ -213,20 +216,34 @@ def get_clearsky_ghi(self):
...
-                da.ceil(len(self.rasterizer.time_index) / cs_ghi.shape[-1])
-            )
-            cs_ghi = da.repeat(cs_ghi, n, axis=2)
+            for year in self.rasterizer.time_index.year.unique():
+                n = (self.rasterizer.time_index.year == year).sum()
+                multi_year.append(cs_ghi[..., :n])
 
-        cs_ghi = cs_ghi[..., : len(self.rasterizer.time_index)]
+        cs_ghi = da.concatenate(multi_year, axis=-1)
+        cs_ghi = cs_ghi[..., :len(self.rasterizer.time_index)]
```
bnb32, line 230: Maybe this would be cleaner if we just matched the doy in ``cs_ghi.time`` with the doy in ``self.rasterizer.time_index`` for each element?

grantbuster: Very good call. Check this out. Will want to test on large timeseries once kestrel is back up but i think this should work!

https://github.com/NREL/sup3r/pull/249/commits/13b11dcc4c09ce5f83b58dca22fb4ff7c531cf3d

bnb32: Yeah that looks nice. I think you can also just store ``cs_ghi[Dimension.TIME]`` right after coarsening https://github.com/NREL/sup3r/blob/2d33c8b32f6f26f2171457170e2e828dc4610cc9/sup3r/preprocessing/data_handlers/nc_cc.py#L207 to get what you have here https://github.com/NREL/sup3r/blob/2d33c8b32f6f26f2171457170e2e828dc4610cc9/sup3r/preprocessing/data_handlers/nc_cc.py#L220

grantbuster: hmmm maybe im misunderstanding but i dont see the benefit of storing the index up at line 207 vs. pulling it at 220 and converting to month-day. I kind of like having everything contained in these few lines that reindex the ti. I should def be using `Dimension.TIME` though. will push that.

bnb32: Yeah, disregard that initial message. At first glance I thought there was some extra unneeded conversion happening. Edited to clarify.

grantbuster: Gotcha! I think I saw the time index is stored as a numpy array so needed the to datetime. doesnt hurt anyway.

bnb32 (approved):

grantbuster:
Just verified this functionality on kestrel with a 10 year timeseries, everything works as expected, merging.

PR #254, opened 2025-01-08
interpolation adjustment: option to include single levels in interpol…

bnb32:
…ation along with isobars. Doing so can sometimes lead to issues if values computed from pressure levels differ from the single level values. This can result in very large alpha values in the linear interpolation routine.

### DIFF
--- a/sup3r/preprocessing/data_handlers/factory.py
+++ b/sup3r/preprocessing/data_handlers/factory.py
@@ -47,6 +47,7 @@ class DataHandler(Deriver):
         self,
         file_paths,
         features='all',
+        load_features='all',
         res_kwargs: Optional[dict] = None,
         chunks: Union[str, Dict[str, int]] = 'auto',
         target: Optional[tuple] = None,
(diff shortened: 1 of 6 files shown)


grantbuster (approved):
@bnb32  are you adding this primarily to ignore the single level values? They were previously included by default? 

The fact that this is such an issue concerns me and makes me think we should be using the single level values (if available) instead of pressure levels.

bnb32:
@grantbuster Yeah I'm using this to ignore single level values when I want to interpolate to hub heights. I found that when I was interpolating temperature to 100m some of the heights in the pressure level data were close to 2m. If I included the single level temperature_2m values I'd end up with two interpolation heights close to 2m (e.g. 2 and 2.001 or something), since the interpolation method uses the two closest heights to 100m, even if they are both below 100m. This gives a very high alpha value in the interpolation scheme which then outputs nonsense when extrapolating this to 100m. It's not possible to use only single levels for temp interpolation, since there's only 2m, and I didn't see this issue with windspeeds (which already have single level values at 10 and 100, for ERA). I could modify the ``interp_kwargs`` to enable only using the single levels though.

bnb32:
@grantbuster I changed the approach here to enable skipping either pressure level variables or single level variables. Instead of adding a bunch of options to ``interp_kwargs`` you can just specify which features to load (which will be used in derivations) using the ``load_features`` kwarg.

PR #255, opened 2025-01-15
need to parse FeatureRegistry so no overlap in kwargs

grantbuster:


bnb32 (commented):
I thought you might run into issues with this kwarg, and it shows that the ``DataHandlerFactory``was just an example of being fancy for the sake of being fancy. Removed this and resolved the kwarg pass through in #256

bnb32 (approved):

PR #258, opened 2025-02-04
Bnb/min padding general

bnb32:
This generalizes the padding fix implemented in #247

### DIFF
--- a/sup3r/pipeline/slicer.py
+++ b/sup3r/pipeline/slicer.py
@@ -656,8 +656,9 @@ class ForwardPassSlicer:
         # modulo(grid_size, fwp_chunk_shape) is less than the padding applied
         # in the first padding layer of the generator
         padded_width = 2 * max_pad + win_stop - win_start
-        too_small = padded_width < min_width
-        if check_boundary and win_stop == max_steps and too_small:
+        is_last_slice = win_stop == max_steps
+        too_small = min_width is not None and padded_width < min_width
+        if check_boundary and is_last_slice and too_small:
             half_width = min_width // 2 + 1
             stop = np.max([half_width, max_pad])
             start = np.max([half_width, max_pad])
--- a/sup3r/pipeline/slicer.py
+++ b/sup3r/pipeline/slicer.py
@@ -59,7 +59,9 @@ class ForwardPassSlicer:
         Minimum width of padded slices, with each element providing the min
         width for the corresponding dimension. e.g. (spatial_1, spatial_2,
         temporal). This is used to make sure generator network input meets the
-        minimum size requirement for padding layers.
+        minimum size requirement for padding layers. e.g. If the generator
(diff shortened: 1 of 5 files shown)


sup3r/pipeline/slicer.py
```diff
@@ -548,19 +566,22 @@ def check_boundary_slice(self, unpadded_slices, cropped_slices, dim):
         lr_slice_stop = unpadded_slices[-1].stop or self.coarse_shape[dim]
 
         # last slice adjustment
-        if 2 * self.spatial_pad + (lr_slice_stop - lr_slice_start) < 4:
+        if (
+            2 * padding + (lr_slice_stop - lr_slice_start)
+            <= self.min_width[dim]
```
grantbuster, line 571: Personally, i find it really helpful to define some of these weird conditionals, both for readability and also to reduce this crazy one-liner. What are these? Just guessing but i'd do:

```
slice_width = lr_slice_stop - lr_slice_start
padded_width = 2 * padding + slice_width
too_small = padded_width <= self.min_width[dim]
if too_small:
    ...
```
And also wait i would think that if we're `padded_width == min_width` we should be okay based on the docstring for "min_width"? Surprised that it's <=

sup3r/pipeline/slicer.py
```diff
@@ -625,14 +652,16 @@ def _get_pad_width(window, max_steps, max_pad, check_boundary=False):
...
         # We add minimum padding to the last slice if the padded window is
         # too small for the generator. This can happen if 2 * spatial_pad +
-        # modulo(grid_size, fwp_chunk_shape) < 4
+        # modulo(grid_size, fwp_chunk_shape) is less than the padding applied
+        # in the first padding layer of the generator
         if (
             check_boundary
             and win_stop == max_steps
-            and (2 * max_pad + win_stop - win_start) < 4
+            and (2 * max_pad + win_stop - win_start) <= min_width
```
grantbuster, line 660: same comment here, i'd recommend breaking this into semantic variables thats easier for future us to read.

sup3r/pipeline/strategy.py
```diff
@@ -253,6 +253,29 @@ def __post_init__(self):
...
+            return pad_width
+        for layer in layers:
+            if hasattr(layer, 'paddings'):
+                new_pw = np.max(layer.paddings, axis=1)[1:-1]
+                if len(new_pw) < 3:
+                    new_pw = (*new_pw, 1)
+                pad_width = [
+                    np.max((new_pw[i], pad_width[i])) for i in range(3)
+                ]
+        return pad_width
```
grantbuster, line 277: This feels like it could be a bit specific to our current model design. What if we migrate to a different padding layer or change the attr in FlexiblePadding? Seems like a pretty rigid tight coupling. Could we just have min_width be an optional user arg that defaults to (1,1,1)?

grantbuster (changes requested):

grantbuster (approved):

PR #272, opened 2025-08-02
Bnb/obs scaling

bnb32:
Added parameter for scaling obs data in obs rasterizer. Small refactor of arguments in ExoDataHandler.

### DIFF
--- a/sup3r/preprocessing/rasterizers/exo.py
+++ b/sup3r/preprocessing/rasterizers/exo.py
@@ -82,8 +82,12 @@ class BaseExoRasterizer(ABC):
     source_handler_kwargs : dict | None
         Any kwargs for initializing the source handler
         (:class:`~sup3r.preprocessing.Loader`).
-    cache_dir : str | './exo_cache'
-        Directory to use for caching rasterized data.
+    cache_dir : str | None
+        Directory to use for caching rasterized data. If None (default) then no
+        data will be cached. If a string is provided then this will be
+        created if it does not exist and the rasterized data will be saved to
+        this directory. This is useful for speeding up subsequent runs of the
+        rasterizer.
     chunks : str | dict
         Dictionary of dimension chunk sizes for returned exo data. e.g.
         {'time': 100, 'south_north': 100, 'west_east': 100}. This can also just
@@ -116,7 +120,7 @@ class BaseExoRasterizer(ABC):
     t_enhance: int = 1
     input_handler_name: Optional[str] = None
     input_handler_kwargs: Optional[dict] = None
-    cache_dir: str = './exo_cache/'
+    cache_dir: Optional[str] = None
     chunks: Optional[Union[str, dict]] = 'auto'
     distance_upper_bound: Optional[int] = None
     fill_nans: bool = True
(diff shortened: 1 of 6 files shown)


sup3r/bias/bias_calc_vortex.py
```diff
@@ -125,8 +125,8 @@ def convert_month_height_tif(self, month, height):
                 }
             )
             ds = ds.isel(band=0).drop_vars('band')
-            ds.to_netcdf(outfile, format='NETCDF4', engine='h5netcdf')
-        return outfile
+            ds.to_netcdf(out_fille, format='NETCDF4', engine='h5netcdf')
+        return out_fille
```
grantbuster, line 129: why the extra L?

sup3r/preprocessing/rasterizers/exo.py
```diff
@@ -383,27 +399,46 @@ def _get_data_2d(self):
         return df[self.feature].values.reshape(self.hr_shape[:-1])
 
     def _get_data_3d(self):
```
grantbuster, line 401: recommend adding docstring describing output shapes/dimensions here.

bnb32 (commented):
> Minor comment but generally i'm a bit nervous about this. Have we figured out why the MADIS obs are negatively biased? How are you determining a scale factor?  

Don't have a complete explanation for the bias of MADIS vs WTK but that's how I'm determining the scale factor (average of wtk(at 10m) / madis over multiple sites). I haven't been using this right now but I think something like this makes sense so that the relationship between near surface and hub height learned by the model is preserved during inference.

grantbuster:
> > Minor comment but generally i'm a bit nervous about this. Have we figured out why the MADIS obs are negatively biased? How are you determining a scale factor?
> 
> Don't have a complete explanation for the bias of MADIS vs WTK but that's how I'm determining the scale factor (average of wtk(at 10m) / madis over multiple sites). I haven't been using this right now but I think something like this makes sense so that the relationship between near surface and hub height learned by the model is preserved during inference.

Well the model will preserve those relationships whether we like it or not, just a matter of bias at surface and consequently at hub height, right? The model should learn to respect and assimilate the obs we give it, but if we don't like the obs then the question is what do we do? Does the WTK have tons of bias at 10m? This fudge factor makes sense but is it scientifically defensible? Would the fudge factor really be constant for all locations and times?

grantbuster:
> > > > Minor comment but generally i'm a bit nervous about this. Have we figured out why the MADIS obs are negatively biased? How are you determining a scale factor?
> > > 
> > > 
> > > Don't have a complete explanation for the bias of MADIS vs WTK but that's how I'm determining the scale factor (average of wtk(at 10m) / madis over multiple sites). I haven't been using this right now but I think something like this makes sense so that the relationship between near surface and hub height learned by the model is preserved during inference.
> > 
> > 
> > Well the model will preserve those relationships whether we like it or not, just a matter of bias at surface and consequently at hub height, right? The model should learn to respect and assimilate the obs we give it, but if we don't like the obs then the question is what do we do? Does the WTK have tons of bias at 10m? This fudge factor makes sense but is it scientifically defensible? Would the fudge factor really be constant for all locations and times?
> 
> The value of near surface ws vs hub height ws is part of that relationship. Either way though, this is just following your suggestion to scale madis to wtk that came up in the discussion on what is the true measurement height for madis. Whether wtk is biased at 10m is unknown, since the comparison would be to madis... Of course the global scale factor is the simplest iteration and it would be better to apply some location / time dependent correction, but that isn't feasible right now.

I hear ya i'm just realizing how hacky this is. Do you know what scale factor you'd use and what kind of delta h that would imply using the power law? If you use this in published data or a manuscript just be really careful, will be hard to defend i think.

grantbuster (approved):

PR #279, opened 2025-09-11
Bnb/exo agg methods

bnb32:
This includes idw, nn, and mean based agg methods for exo data. Also includes some minor batching speed ups - turning off tensorboard logging is a big one.

### DIFF
--- a/tests/rasterizers/test_exo.py
+++ b/tests/rasterizers/test_exo.py
@@ -366,8 +366,7 @@ def test_obs_agg(s_enhance, with_nans):
             source_files=f'{td}/hr.nc',
             feature='u_10m_obs',
             s_enhance=s_enhance,
-            t_enhance=1,
-            agg_method='mean',
+            t_enhance=1
         )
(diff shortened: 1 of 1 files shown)


sup3r/preprocessing/rasterizers/exo.py
```diff
@@ -413,27 +460,39 @@ def _get_data_3d(self):
...
+        elif self.agg_method == 'nn':
+            func = self._nn_fill
+        elif self.agg_method == 'mean':
+            func = self._mean_fill
+        else:
+            raise ValueError(
+                f'Unknown aggregation method: {self.agg_method}. '
+                'Must be one of "idw", "nn", or "mean".'
+            )
+        logger.info('Using {} aggregation method'.format(self.agg_method))
```
bnb32, line 474: Not sure if its worth keeping nn and idw options.

bnb32: removed all but the mean agg method

PR #293, opened 2025-12-22
time_feature added for temporal encoding. tests added for basic featu…

bnb32:
hour of day and hour of year encoding features added; based on SatMAE positional encoding.

### DIFF
--- a/sup3r/preprocessing/derivers/methods.py
+++ b/sup3r/preprocessing/derivers/methods.py
@@ -431,10 +431,16 @@ class SpatioTemporalEncoding(DerivedFeature):
     .. [cong2022] Cong, Yezhen, et al. "Satmae: Pre-training transformers for
     temporal and multi-spectral satellite imagery." Advances in Neural
     Information Processing Systems 35 (2022): 197-211.
+
+    Note
+    ----
+    We use the referenced paper for inspiration but modify the argument of the
+    trig functions to be 2πi * k / d so the periodicity matches the underlying
+    feature (e.g. hour of day repeats every 24 hours).
     """
 
     @classmethod
-    def _compute(cls, data, k, k_dims, d, i=1, omega=10000):
+    def _compute(cls, data, k, k_dims, d, i=0):
         """Compute method for encoding.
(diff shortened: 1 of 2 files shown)


bnb32:
> I'm plotting this up and honestly i'm kind of confused why this is so widely used. Maybe it makes sense if you can afford a large number of frequencies but if we only want 1-2 frequencies wouldn't it make more sense to explicitly define the periodicity like this?
> 
> ```
>         k = k * (i + 1) * (np.pi / d)
>         k = np.sin(k) if i % 2 == 0 else np.cos(k)
> ```
> 
> Also, didn't we agree to do SecondOfDay/Year to preemptively protect against the sub-hourly sup3r?

I had basically the same thought. So strange not to just use something like a wavenumber 2pi / d instead of this weird thing with omega. But this is supposed to be sota so what do I know lol. And yeah, forgot to do seconds. Thanks.

bnb32:
> > > I'm plotting this up and honestly i'm kind of confused why this is so widely used. Maybe it makes sense if you can afford a large number of frequencies but if we only want 1-2 frequencies wouldn't it make more sense to explicitly define the periodicity like this?
> > > ```
> > >         k = k * (i + 1) * (np.pi / d)
> > >         k = np.sin(k) if i % 2 == 0 else np.cos(k)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also, didn't we agree to do SecondOfDay/Year to preemptively protect against the sub-hourly sup3r?
> > 
> > 
> > I had basically the same thought. So strange not to just use something like a wavenumber 2pi / d instead of this weird thing with omega. But this is supposed to be sota so what do I know lol. And yeah, forgot to do seconds. Thanks.
> 
> I vote we do what's more intuitive to us esp. given that we have limited bandwidth for these channels in our model... I think this formulation was introduced by the original transformer paper that did not have periodicity.

Sounds good to me.

grantbuster (approved):

PR #297, opened 2026-02-20
lat / lon sinusoidal encoding

bnb32:


ppinchuk:
> @grantbuster I'll merge this in tomorrow if you have no objections. @ppinchuk, I just added you so you'd be aware of this positional encoding for the geothermal work.

Thanks, I was wondering about this as I was going through the training script over the weekend. Good to know that I wasn't just crazy and couldn't find it in the code :)

PR #301, opened 2026-02-27
feat: custom loss with specific input features.

bnb32:


sup3r/utilities/loss_metrics.py
```diff
@@ -9,7 +9,16 @@
 from tensorflow.keras.losses import MeanAbsoluteError, MeanSquaredError
 
 
-def _derivative(x, axis=1):
+class PhysicsBasedLoss(tf.keras.losses.Loss):
+    """Base class for physics-based loss metrics. This is meant to be used as a
+    base class for loss metrics that require specific input features."""
+
+    def __init__(self, input_features='all'):
```
ppinchuk: So the order of `input_features` is very important right? If I'm reading this correctly, it determines the order that the features appear in the tensor. Is that right? If so, might be worth documenting to lower the barrier to entry for users like me 😄

ppinchuk (commented):
Freaking cool! Did you want me to work on adding the geothermal losses in this PR or a subsequent one?

bnb32:
> Freaking cool! Did you want me to work on adding the geothermal losses in this PR or a subsequent one?

Thanks! Btw, I thought the assertion vs raise suggestion was good. TIL assertions could be removed with optimization flags. 

Subsequent PR works for me.

